PR #16, opened 2015-04-23
Removed id from Object's immutable fields

jamespstrachan:
Opening PR for feedback/clarification.

Objects retrieved by use of their reference (rather than their id) have this reference set their id parameter. This borks subsequent API calls made by the object, often yielding a `The requested item doesn't exist (404)` error.

I am unsure what the side-effects of making this id mutable may be. I can't see any obvious consequences from examining the codebase, and it does make retrieved objects as functional as newly created ones.

Am I missing anything? Happy to maintain my own fork if this works but only for my specific purposes, but would like to know of potential unseen consequences of the change.


att14:
This is fixed in version 3.0.0

PR #17, opened 2015-05-11
Added timeout to EasyPost config

gsinkin:
Added timeout to EasyPost config.
Increased request timeout from 80s to 120s.
Bumped version to 2.1.1


briedis:
Please merge this is, this hardcoded value is real pain in the ass, because we had to manually edit this libraries source to make it work.

briedis:
@gsinkin, you should probably move the connection timeout too as a changable value.

gsinkin:
You got it. Having one of our PHP devs give it a once over and in she goes.

sawyer:
We're going to go a different way and remove those values from being explicitly set in c43f3c6a4ba30e0e9a5ac0a259061062c78957bb.

We consider the default values acceptable, and are in full support of a community fork if there's a solid need for setting timeout values on a per request basis.

PR #24, opened 2015-12-08
better composer dependency and spelling fixes

browner12:
we should be encouraging users to require stable released versions, not `dev-master`. that is bad practice.  as long as you are following semantic versioning (which it appears you are) we can say we want anything in the major 2 version.

also a spelling fix

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,17 +25,17 @@ Create or add the following to composer.json in your project root:
         }
     ],
     "require": {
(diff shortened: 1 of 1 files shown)


browner12:
could someone please look at this? it should be a pretty straight forward merge

browner12:
this is a pretty straightforward PR. could someone *please* look at it

PR #25, opened 2015-12-08
increase php version requirement

browner12:
v5.3 and v5.4 are EOL. v5.5 is the current minimum supported version.

i understand you may have some concerns bumping it all the way to 5.5,
but if you’re not comfortable with that please at least go to 5.4

5.3 has been EOL for over a year and has many known security issues.

bumping this is a good idea not only for security, but also gives the
package the ability to start taking advantage of newer features

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,7 @@
     "require": {
         "ext-curl": "*",
         "ext-json": "*",
-        "php": ">=5.3.0"
+        "php": ">=5.5.0"
     },
     "require-dev": {
         "phpunit/phpunit": "3.7.*"


dangrossman:
You target the platforms your customers are on, not the one that has the shiny features your developers want. A request like this should be backed by a survey of budget shared hosting services' current deployed PHP versions. Very popular distributions among hosting services, like RHEL/CentOS, are still on PHP 5.3, and backport security patches to that version as needed.

2toginc:
I agree with @dangrossman, for what that's worth.

You can always deploy on a shiny version, but this package does not need any of that.  The easypost api is well written, and in most cases you can make do with your own "raw" calls to it, completely bypassing this library.

It's a shame resources are used to maintain these "api layer libs" in all these various languages when we could just be using GET,POST,DELETE like nature intended -- and this is what you will have to do anyway if you are not one of the 6 or so currently supported, and don't want to roll the dice with a "community supported" lib /rant :-)

If the library itself has a security issue on some supported platform (deprecated or otherwise), that could be addressed, but I don't think that's the case here.

Thanks for listening.

browner12:
i completely understand where you are coming from @dangrossman . however, I believe package maintainers have a responsibility to help move the ecosystem forward as well.  like I said, maybe 5.5 was too much of a leap. maybe 5.4 would be more appropriate.

for what it's worth, http://phpversions.info/shared-hosting/

as you can see, there are only 3 shared hosting providers listed that don't offer anything above 5.3.

or, maybe, you start a v3 release that goes to 5.5. this would mean people who couldn't (or choose not to) upgrade their php version can stay on the v2 release, and the rest of us can move forward. then v2 only gets security fixes, and v3 can start utilizing new features. (and i'm not saying use new features simply because they exist, rather use them if they can help us improve)

>It's a shame resources are used to maintain these "api layer libs" in all these various languages when we could just be using GET,POST,DELETE like nature intended

as for these "api layer libs", there's a reason they exist. yes, of course, we could all just write CURL requests, but we write containers to encapsulate code, and make it incredibly easy and fluent for the end user.

>If the library itself has a security issue on some supported platform (deprecated or otherwise), that could be addressed, but I don't think that's the case here.

php5.3 is known to have certain security issues, and because it is EOL, it won't receive patches to any new security issues.  I know that has nothing to do with this library, but still something to consider.

thanks.

browner12:
any more discussion on this?

browner12:
alright, this seems like a lost cause for now. i would suggest we look at again on the next major version change, and possibly start targeting a newer php version.

PR #26, opened 2015-12-09
add docblocks

browner12:
docblocks are a great way to document code.

they are also great for IDEs and helping them with their autocomplete
and code analysis.

i did the best i could describing methods. most of them are self explanatory, and the ones I wasn’t sure on I put very general descriptions in. while it may be beneficial for one of the maintainers to go through again afterwards and put in better descriptions, this is a good starting point.

### DIFF
--- a/lib/EasyPost/Address.php
+++ b/lib/EasyPost/Address.php
@@ -4,21 +4,47 @@ namespace EasyPost;
 
 class Address extends Resource
 {
+    /**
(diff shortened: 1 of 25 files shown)


browner12:
could someone please look at this? it would be great to have docblocks in the code to help with code completion.

browner12:
bueller? docblocks = good

PR #29, opened 2016-01-06
Docblocks

victoryftw:


browner12:
glad to help! and thanks for merging, just an FYI, you mistyped my name.

also, when did version 3 come out? is it the current stable version? because if so we should update the readme again to reference that version.

victoryftw:
Thanks for letting me know about the misspelling! I corrected it now.

I also bumped the version in the readme again to reflect that we are now on version 3.

browner12:
according to the changelog, there were no breaking changes. why did you guys bump to v3 instead of a minor or a patch?

victoryftw:
That was originally our plan; however, we realized that the change could in fact be a breaking change if any of our users relied on the old (incorrect) behavior. Since there was a value being returned in the ID field in those cases, just not the correct value, we took the safe approach and made a major version bump just in case anyone was keying off that field.

browner12:
awesome, thanks for the replies

PR #37, opened 2016-06-02
Resource is a reserved word in PHP 7

ccovey:
Resource is a reserved word in PHP 7. This changes all instances of the name to ObjectResource.


jsuggs:
:+1:

victoryftw:
Thanks for this! We decided to go with "EasypostResource" over "ObjectResource", but this was a big help! I'll release the new version shortly.

PR #50, opened 2017-06-28
added standalone rating

algreen11:
Can now create rating calls from php (the endpoint is buggy sometimes)


Justintime50:
Standalone rating is no longer supported. Thank you for your efforts on this PR!

PR #51, opened 2017-08-17
Remove exception

NGPrakash-Techshetra:


att14:
Can you provide an explanation of why you are doing this?

PR #53, opened 2017-11-16
Add optional CURL timeouts setting

mattjanssen:
This PR addresses Issue https://github.com/EasyPost/easypost-php/issues/54

This is a backwards-compatible change that adds the `EasyPost::setConnectTimeout()` and  `EasyPost::setResponseTimeout()` methods to allowing setting the number of milliseconds to wait for a connection and response, respectively,

These are needed because the default curl behavior is to wait indefinitely, which can cause services to stall under load.

**Usage**
If `connectTimeout` is set to `1000` it will wait up to one (1) second to connect to EasyPost. If this is set to zero (0) or null, then the default PHP curl behavior is maintained. Thus there is no BC break, because the defaults for these settings are both null.

Likewise, if `responseTimeout` is set to `5000` it will wait up to five (5) seconds, after the connection is established, to get a complete response from EasyPost. If this is set to zero (0) or null, then the default PHP curl behavior is maintained.

The two timeouts are kept separate because we should be able to expect a connection within a fairly short timeframe, but may need to wait longer for a response.

### DIFF
--- a/lib/EasyPost/EasyPost.php
+++ b/lib/EasyPost/EasyPost.php
@@ -20,6 +20,24 @@ abstract class EasyPost
     public static $apiVersion = "2";
 
     /**
+     * Time in milliseconds to wait for a connection
+     *
+     * Zero or null means no timeout.
+     *
+     * @var int|null
+     */
+    public static $connectTimeout;
+
+    /**
+     * Time in milliseconds to wait for a response
+     *
+     * Zero or null means no timeout.
+     *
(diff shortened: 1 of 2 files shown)


kcaporaso:
+1 This would be very useful, thanks @mattjanssen

victoryftw:
Sorry its taken us so long to respond to this; it seems like a quality improvement for sure! We'll release an updated version (3.4.1) with this change shortly.

PR #65, opened 2019-05-08
Fix #63 - Create Shipment Report always returns current date

### LINKED ISSUE #63: Create Shipment Report always returns current date - 31 days report
Hi,

We are using your library in our service package and we started adding the reports section.

We want to be able to create a new report (or fetch it if already exists) with start/end date params being passed.

We noticed that even if our input on was correct

```
$request = [
    'start_date' => $startDate,
    'end_date' => $endDate,
    'type' => $type,
    'send_mail' => false,
];

$report = Report::create($request);
```

We always got a report for current date - 31 days.

After stepping into the main library during debugging, we noticed that the whole array is transformed to

`array('report' => $request)`

which is causing the request to fail for the intented purpose (get report for a specific range) as the post params are wrong for the API.

You could refactor for a solution in various ways,

such as refactor the Report resource class to set the date ranges and other params back into the first level of the array and keep the type only on the level 2 `report` array.

Or, please update the docs with an example that would work for the purpose as the below (checked and worked)

```
$request = [
    'report' => [
        'type' => $type
    ],
    'start_date' => $startDate,
    'end_date' => $endDate,
    'send_mail' => false,
];
```

As far as we can understand, this should affect all other reports.

If we have the time, we could work on a pull request

Many Thanks

pascalroget:


jontsai:
Went with the approach in #78

PR #67, opened 2019-09-20
Update EasyPostObject.php

khvalov:
Providing access to private variable values. Case: accessing parcel settings in Rates object.

### DIFF
--- a/lib/EasyPost/EasyPostObject.php
+++ b/lib/EasyPost/EasyPostObject.php
@@ -135,6 +135,17 @@ class EasyPostObject implements \ArrayAccess, \Iterator
             }
         }
     }
+    
+    /**
+     * Getter to recieve private defined variables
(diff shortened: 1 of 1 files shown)


Justintime50:
@khvalov can you rebase against master so we can take another look?

Can you also provide a simple code example so I can see the use-case in action?

Justintime50:
You mention `accessing parcel settings in Rates object.` - I'm not sure I follow. Parcel parameters and rate parameters are separate as they are different objects. If you want to access something, why not do something such as:

```php
parcel = \EasyPost\Parcel::retrieve('prcl_123...')

echo parcel['weight']
```

I could be misunderstanding your request but I can't initially see a reason to add this functionality. If I've misunderstood, please feel free to leave some additional details on this PR and we can revisit this.

Issue #69, opened 2019-12-19
Setup Travis CI


Justintime50:
Other EasyPost libraries use Travis CI for testing. We should incorporate a PHP linter and PHP unit testing via Travis CI.

Example of .travis.yml:

```
language: php

php:
  - 7.2
  - 7.3
  - 7.4

install:
  - composer self-update
  - composer install -q --no-ansi --no-interaction --no-scripts --no-suggest --no-progress --prefer-dist

script:
  - ./bin/phpunit # Run unit tests
  - ./vendor/bin/phplint ./src --exclude=vendor # do some php linting, will need to install a linter
```

PR #70, opened 2019-12-20
Fixes for v3.4.2

### LINKED ISSUE #61: "Undefined variable: v" when deleting carrier account.
I'm getting an exception, "Undefined variable: v", when trying to do this:

`\EasyPost\CarrierAccount::retrieve("ca_...")->delete();`

It looks like there is a bug in the __unset method of EasyPostObject.php. On line 123 the value of $v is being assigned. But I don't see $v being defined anywhere in that method.

https://github.com/EasyPost/easypost-php/blob/517b6205c9316d67da2241345fa9d52b87ddedf4/lib/EasyPost/EasyPostObject.php#L123

### LINKED ISSUE #68: Write Unit Testing - Switch from PHPUnit 3 to PHPUnit 7/8
There is a single incomplete unit test in this library and it's using PHPUnit 3 which is incredibly deprecated. We should upgrade to PHPUnit 7/8 which is compatible with PHP 7.x and write unit tests for more than just the Address object.

Justintime50:
There was some decent cleanup that the project needed along with testing and linting. This update will pave the way for future work I'll be doing such as setting up unit tests and fixing bugs, TODO's, and other items that have been neglected.

One note to consider is we can't test 5.6 with PHPUnit 8 in Travis even though we may still have customers rocking PHP 5.x; PHPUnit 8 only supports back to PHP 7.2. The previous version of PHPUnit used in this project was released in 2013 and dropped support in 2016 so it's no longer feasible.

Fixes:
* Added PHP linting
* Fixed PHP coding standards - bad elseif statements and missing brackets for if statements among others
* Closes #61 - Undefined variable on carrier deletion
* Closes #68 and #69 - Add unit tests, update PHPUnit, and setup Travis CI
* Cleaned up README extensively


lib/EasyPost/Fee.php
```diff
@@ -4,5 +4,5 @@
 
 class Fee extends EasypostResource
 {
-    
+    //
```
jontsai: were these comments autogenerated?

lib/EasyPost/Requestor.php
```diff
@@ -38,7 +38,7 @@ public static function apiUrl($url = '')
      */
     public static function utf8($value)
     {
-        if (is_string($value) && mb_detect_encoding($value, "UTF-8", TRUE) != "UTF-8") {
+        if (is_string($value) && mb_detect_encoding($value, "UTF-8", true) != "UTF-8") {
```
jontsai: was this autogenerated?

lib/EasyPost/Fee.php
Justintime50: I believe I added these, it's an empty class and have seen blank comments in other projects when a class is empty. Can be removed.

lib/EasyPost/Requestor.php
Justintime50: This change would have come about from linting.

Justintime50:
@jontsai So are you suggesting that no linting changes be made at all or just breaking up the fixes into separate PRs such as 1) README updates, 2) add linter and Travis CI, 3) add unit tests, 4) Make linting fixes, etc?

Justintime50:
Closing in favor of breaking these up into separate PR’s.

PR #74, opened 2020-02-11
Updated test suite, added more basic unit tests

### LINKED ISSUE #68: Write Unit Testing - Switch from PHPUnit 3 to PHPUnit 7/8

Justintime50:
Updated PHPUnit from v3.x to v8.x, tweaked single previously included unit test and added a Parcel and shipment unit test. More to come but this is ready to merge if approved.

Closes #68

### DIFF
--- a/test/EasyPost/Test/AddressTest.php
+++ b/test/EasyPost/Test/AddressTest.php
@@ -7,11 +7,16 @@ use EasyPost\EasyPost;
 
 EasyPost::setApiKey('cueqNZUb3ldeWTNX7MU3Mel8UXtaAMUi');
 
-class AddressTest extends \PHPUnit_Framework_TestCase
+class AddressTest extends \PHPUnit\Framework\TestCase
 {
 
     // TODO: set up tests for exceptions and error codes
 
+    /**
+     * Test the creation of an address
+     *
+     * @return void
+     */
     public function testCreate()
     {
         $address_params = array("street1" => "388 Townsend St",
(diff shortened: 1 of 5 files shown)


test/EasyPost/Test/ParcelTest.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+namespace EasyPost\Test;
+
+use EasyPost\Parcel;
+use EasyPost\EasyPost;
+
+EasyPost::setApiKey('cueqNZUb3ldeWTNX7MU3Mel8UXtaAMUi');
```
jontsai: is this a testing API key? it probably shouldn't be checked into code, and be passed as an environment variable.

Something like this: https://docs.travis-ci.com/user/environment-variables/
And I think this: https://www.php.net/manual/en/function.getenv.php

test/EasyPost/Test/ShipmentTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+namespace EasyPost\Test;
+
+use EasyPost\Shipment;
+use EasyPost\EasyPost;
+
+EasyPost::setApiKey('cueqNZUb3ldeWTNX7MU3Mel8UXtaAMUi');
```
jontsai: don't check in API keys to code

test/EasyPost/Test/AddressTest.php
```diff
@@ -7,67 +7,47 @@
 
 EasyPost::setApiKey('cueqNZUb3ldeWTNX7MU3Mel8UXtaAMUi');
```
jontsai: load the API key using an environment variable

test/EasyPost/Test/ParcelTest.php
att14: This is how all of the client libraries do it as of today. I think this is outside the scope of this change.

Justintime50: @jontsai I agree with @att14 as all libraries are setup this way currently, but I also agree with you that moving forward we should set these up as environment variables instead!

jontsai: sgtm

PR #77, opened 2020-02-19
Install info & Travis badge fixes on README

Justintime50:
Fixed the Travis badge from `.org to .com` and made some additional README clarifications after looking through everything again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,30 +1,30 @@
 # EasyPost PHP Client Library
 
-[![Build Status](https://travis-ci.org/EasyPost/easypost-php.svg?branch=master)](https://travis-ci.org/EasyPost/easypost-php)
+[![Build Status](https://travis-ci.com/EasyPost/easypost-php.svg?branch=master)](https://travis-ci.com/EasyPost/easypost-php)
 
 EasyPost is a simple shipping API. You can sign up for an account at https://easypost.com.
 
 ## Installation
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,58 +1,54 @@
...
-require_once("path/to/vendor/autoload.php");
-\EasyPost\EasyPost::setApiKey('cueqNZUb3ldeWTNX7MU3Mel8UXtaAMUi');
+require_once("/path/to/vendor/autoload.php");
+use \EasyPost\EasyPost;
+use \EasyPost\Address;
+use \EasyPost\Parcel;
+use \EasyPost\Shipment;
+EasyPost::setApiKey('cueqNZUb3ldeWTNX7MU3Mel8UXtaAMUi');
 
-$to_address = \EasyPost\Address::create(
```
jontsai: I think keeping the full namespace is better. If people are skimming examples and just want to get code working ASAP as quickly as possible, they are probably going to copy/paste select portions. The `use \EasyPost\EasyPost;` etc adds additional elements that aren't significant improvements, IMO, at the potential expense of reduced clarity.

README.md
```diff
@@ -1,58 +1,54 @@
...
 
 ```shell
 composer require easypost/easypost-php
 ```
 
-Require dependencies:
+Then, require the autoloader:
 
 ```php
 require_once("/path/to/vendor/autoload.php");
```
jontsai: Should this be `/path/to/vendor/easypost/autoload.php` ?

Justintime50: Yes, it would probably be clearer with `/easypost` added. I'm assuming whoever initially wrote it wrote it as from the library itself which wouldn't include `/easypost`. I can add this in.

README.md
Justintime50: Fair enough! I'll revert this.

PR #78, opened 2020-03-07
Fixed create report feature and added unit test

Justintime50:
Similar to our Python client library, the reports response is being double wrapped when it shouldn't. I've corrected that issue here and written a unit test.

For whatever reason, I cannot get the unit test of `assertInstanceOf` to pass for the `Refund` report... the problem is alluding me as the code is the exact same across all four reports. I'll need some help fixing that unit test before this will pass.

### DIFF
--- a/lib/EasyPost/Report.php
+++ b/lib/EasyPost/Report.php
@@ -53,16 +53,10 @@ class Report extends EasypostResource
      */
     public static function create($params = null, $apiKey = null)
     {
-        if (!isset($params['report']) || !is_array($params['report'])) {
-            $clone = $params;
-            unset($params);
-            $params['report'] = $clone;
-        }
-
-        if (!isset($params['report']['type'])) {
+        if (!isset($params['type'])) {
             throw new Error('Undetermined Report Type');
         } else {
-            $type = $params['report']['type'];
+            $type = $params['type'];
 
             self::_validate('create', $params, $apiKey);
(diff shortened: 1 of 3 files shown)


Justintime50:
@jontsai I've fixed the unit test, there was a missing prefix on the EasyPost object for the Refund Report. Should be all set!

Justintime50:
Closes #63 & is more robust than #65 PR. Could be closed once this is merged.

lib/EasyPost/Report.php
```diff
@@ -53,16 +53,16 @@ public static function all($params = null, $apiKey = null)
      */
     public static function create($params = null, $apiKey = null)
     {
-        if (!isset($params['report']) || !is_array($params['report'])) {
+        if (!isset($params) || !is_array($params)) {
```
jontsai: I'm not sure what the benefit of keeping this block is anymore. We're unsetting and re-setting `$params`, so I believe this block could be deleted altogether

Justintime50: I agree, strange setup; however, the rest of the library is also setup like this. I could remove it if necessary.

jontsai: It was necessary before because it was checking on the value of a nested key within `$params`. Now, it makes no sense when checking the top-level variable.

Justintime50: Sounds good! I have removed this.

PR #86, opened 2020-05-22
Added missing event retrieve methods

Justintime50:
Added the missing `retrieve` and `all` event methods.


Justintime50:
@att14 I plan to do a full VCR unit test PR separately as PHP doesn't really have unit tests yet.

PR #87, opened 2020-05-22
Fixes the bug not allowing for carrier deletion

### LINKED ISSUE #61: "Undefined variable: v" when deleting carrier account.

Justintime50:
Carriers cannot be deleted currently. This fixes the error found in #61 

Closes #61 
Closes #64 (difference here is no changes made to CHANGELOG as we'll do a release PR after)

### DIFF
--- a/lib/EasyPost/EasyPostObject.php
+++ b/lib/EasyPost/EasyPostObject.php
@@ -84,6 +84,7 @@ class EasyPostObject implements \ArrayAccess, \Iterator
         $i = 0;
         $current = $this;
         $param = array($k => $v);
+
         while (true && $i < 99) {
             if (!is_null($current->_parent)) {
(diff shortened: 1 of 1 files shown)


Justintime50:
@att14 I created a child account to test then deleted it with this fix in place. Let me know how else I can show that work here (may be difficult to do).

PR #88, opened 2020-05-22
Removed unpaginated "all" methods

Justintime50:
These 3 endpoints are not paginated and as such, should be removed due to low usage and performance.

### DIFF
--- a/lib/EasyPost/CustomsInfo.php
+++ b/lib/EasyPost/CustomsInfo.php
@@ -17,18 +17,6 @@ class CustomsInfo extends EasypostResource
     }
 
     /**
-     * retrieve all customs info
-     *
-     * @param mixed $params
-     * @param string $apiKey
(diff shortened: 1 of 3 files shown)


lib/EasyPost/CustomsInfo.php
```diff
@@ -17,16 +17,8 @@ public static function retrieve($id, $apiKey = null)
     }
 
     /**
-     * retrieve all customs info
-     *
-     * @param mixed $params
-     * @param string $apiKey
-     * @return mixed
+     * All CustomsInfo not implemented
      */
```
att14, line 21: Are these comments necessary?

Justintime50: Not necessarily. They could be removed.

Justintime50:
@att14 removed the comments in question.

PR #89, opened 2020-05-22
Removed Old/Unusable Methods (Code Cleanup)

Justintime50:
There were many unusable methods or old methods that are removed here to cleanup the project. For instance, almost no EasyPost record can be updated (saved) so those were removed except for users and webhooks which can be updated. Additionally, removed some old label methods that aren't present in other CL's (discussed with @att14 ) as well as removed the `container` and `item` objects as these are not currently a thing and aren't present anywhere else.


att14:
I thought the idea behind `save` was that someone could create an instance of an object and the call save on it, rather than using the static method `create`.

Justintime50:
@att14 I'll play with that and get back to you. I may have been mistaken initially? However, `save` uses the HTTP `put` method which is only listed in our docs for endpoints that you can actually update (users and webhooks). I've never used it to create records although it could be done right?

att14:
Ya, if it uses `PUT` then this change is correct.

Justintime50:
My apologies, it uses `patch` which I thought while writing but `put` came out:

https://github.com/EasyPost/easypost-php/blob/master/lib/EasyPost/EasypostResource.php#L151-163

PR #97, opened 2021-01-08
add github action for ci

Roguelazer:
once these are passing I'll remove the Travis integration


Justintime50:
Can we test against PHP 8 as well for kicks and giggles?

Justintime50 (approved):
Very nice, PHP 5 needs to die...

PR #103, opened 2021-05-11
fix(lint): Swaps linter and lints project

Justintime50:
This PR:
* Swaps the linter
* Lints the entire project
* Adds missing composer commands for convenient running of `lint`, `fix`, and `test`
* Checks in the missing (and ignored) `composer.lock` file so all devs get the same set of dependencies
* Pins `phpunit/php-token-stream` to `v3` to fix build on `PHP 7.2`
* Removes the API key from the README

About a year ago, I introduced a couple dev dependencies that we should remove in favor of better supported and more mainstream alternatives.

The dependencies initially added are `php-cs-fixer` which is a linter/formatter that will automatically correct code. This is too invasive (without an accompanying linter) as it assumes the style on our behalf. There was also `phplint` which does the absolute bare minimum (only checks syntax errors but doesn't check standard style corrections).

We should replace these two dependencies for a single dependency for the following reasons: 
* `php_codesniffer` is a true linter (and fast) that gives useful feedback in an easy-to-read format whereas `phplint` was slower and the output to console wasn't as verbose
* It includes both a linter and a lint corrector in one (different CLI usage) allowing us to remove a dependency
* [phplint](https://github.com/overtrue/phplint) (old) only has 669 stars and is not committed to frequently (last year aside from a single commit this year)
* [php_codesniffer](https://github.com/squizlabs/PHP_CodeSniffer) (new) has 8.6k stars and has been updated within the last week. It's also used in various large frameworks and projects for PHP
* `php_codesniffer` lints against the most recent [PSR-12](https://www.php-fig.org/psr/psr-12/) standard which is the "gold standard" for PHP development. The old `phplint` did not lint against this standard (as is shown by how many linting errors were corrected in this PR)

Replacing the linter will speed up development, improve code readability and errors, and reduce the development overhead of a dependency.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,7 @@ jobs:
       - name: install dependencies
         run: composer install --no-ansi --no-interaction --no-scripts --no-progress --prefer-dist
       - name: lint
-        run: ./bin/phpcs
+        run: composer lint
   run-tests:
     runs-on: ubuntu-18.04
     strategy:
@@ -45,4 +45,4 @@ jobs:
       - name: install dependencies
         run: composer install --no-ansi --no-interaction --no-scripts --no-progress --prefer-dist
       - name: test with phpunit on ${{ matrix.phpversion }}
-        run: ./bin/phpunit
+        run: composer test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,8 +1,8 @@
-name: 'CI'
+name: "CI"
 
 on:
   push:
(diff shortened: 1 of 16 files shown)


jontsai:
Could we include some additional links that speak more to the pros vs cons of each PHP linter library, and why this is better?

I'm not opposed to making the switch; I would just like to see some better documentation for posterity and some more rigorous justification rather than anecdotal evidence.

ajnassar (approved):
LGTM

jontsai (approved):

PR #109, opened 2021-09-07
feat: adds VCR support to tests

Justintime50:
This PR adds long overdue `VCR` support to the PHP client library. This carries multiple benefits:

1. Consistent tests. We can now replay HTTP responses instead of hitting the live API on every test run
2. We can remove the poorly constructed "random date" logic in the reports tests which is now much less flakey
3. We can assert real values for items such as SmartRate integers on delivery days because we can return the same value each time

This PR includes another library called `php-vcr-sanitizer` that I personally contributed to that redacts items such as the `Authorization` header so we don't expose API keys in the cassettes. This library is needed because `php-vcr` is not privacy conscious out of the box.

The hardcoded API Key was removed from tests and replaced with `getenv` to grab the `API_KEY` environment variable. Documentation and CI were updated appropriately.

We also remove testing on PHP 7.2 as it's EOL and bump PHP Unit to v9 to better support newer versions of PHP and as this also drops support for 7.2.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       - name: set up php
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.4
+          php-version: 8.0
       - name: install composer
         run: composer self-update
       - name: install dependencies
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        phpversion: [7.2, 7.3, 7.4, 8.0]
+        phpversion: [7.3, 7.4, 8.0]
     steps:
       - uses: actions/checkout@v2
       - name: set up php
@@ -45,4 +45,4 @@ jobs:
       - name: install dependencies
         run: composer install --no-ansi --no-interaction --no-scripts --no-progress --prefer-dist
       - name: test with phpunit on ${{ matrix.phpversion }}
-        run: composer test
+        run: API_KEY=123 composer test
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ EasyPost is a simple shipping API. You can sign up for an account at https://eas
 
 ## Installation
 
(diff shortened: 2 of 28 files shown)


.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     runs-on: ubuntu-18.04
     strategy:
       matrix:
-        phpversion: [7.2, 7.3, 7.4, 8.0]
+        phpversion: [7.3, 7.4, 8.0]
```
jontsai, line 27: # Reviewer Note

PHP 7.2 has been EOL since Nov 2020

See:
- https://www.php.net/eol.php
- https://www.php.net/supported-versions.php

README.md
```diff
@@ -103,7 +103,7 @@ Up-to-date documentation can be found at: https://www.easypost.com/docs.
 
 ## Testing & Development
 
-***NOTE: Unit tests only work with PHP 7.2^.***
+***NOTE: Unit tests only work with PHP >=7.2 and <8.0.***
```
jontsai:
```suggestion
***NOTE: Unit tests only work with PHP >=7.3 and <8.0.***
```

composer.json
```diff
@@ -1,43 +1,45 @@
...
+      "homepage": "https://www.easypost.com"
+    }
+  ],
+  "require": {
+    "ext-curl": "*",
+    "ext-json": "*",
+    "php": ">=5.3.0"
+  },
+  "require-dev": {
+    "allejo/php-vcr-sanitizer": "^1.0.9",
```
jontsai, line 25: # Reviewer Note

These were the only libraries that were changed. The other line changes are just whitespace/indentation changes (can verify by appending `?w=1` to the review URL to hide whitespace-only changes)

test/bootstrap.php
```diff
@@ -0,0 +1,21 @@
+<?php
+
+use VCR\VCR;
+use allejo\VCR\VCRCleaner;
+
+if (!file_exists('test/cassettes')) {
+    mkdir('test/cassettes', 0777, true);
```
jontsai: Is `777` potentially too permissive? Should would we do something like `0755`?

Reference: https://en.wikipedia.org/wiki/Chmod

jontsai (changes requested):

jontsai (approved):
LGTM

ajnassar (approved):

PR #115, opened 2021-10-08
fix: properly handle empty objects and lists

### LINKED ISSUE #114: Document 4.0 breaking changes
I ran into an issue where

```php
\EasyPost\Shipment::create[
  'from_address' => $fromAddress,
  'to_address' => $toAddress,
  'parcel' => $parcel,
  'options' => [],
]);
```

with empty `options` now results in `EasyPost\Error: The provided options are not valid.`. This was not the case in 3.x using form encoding.

Justintime50:
If you pass an empty array or object with the new JSON encoding, it doesn't get handled properly at the moment (see #114).

This PR will filter out null or empty objects and lists. I added various tests that cover:

1. Empty lists 
2. Empty objects 
3. Null strings
4. Null objects
5. Tax identifiers (which behave oddly if not done correctly, this is the perfect example to test to ensure we did this JSON encoding thing right)
6. Customs Items (which behave opposite of tax identifiers and are another perfect item to test)

As a bonus, I added docstrings to each function in `Requestor` because it's difficult to trace. I hope we'll follow this up with additional fixes and clarifying PRs surrounding this logic in the future.

There may still need to be some tweaking if I've missed some odd edge case but feel pretty good about this.

Closes #114


test/EasyPost/Test/ShipmentTest.php
```diff
@@ -156,4 +156,54 @@ public function testSmartrate()
         $this->assertEquals($smartrates[0]['time_in_transit']['percentile_97'], 3);
         $this->assertEquals($smartrates[0]['time_in_transit']['percentile_99'], 5);
     }
+
+    /**
+     * Test the creation of a Shipment with empty or null objects and arrays
+     *
+     * @return Shipment
+     */
+    public function testCreateEmptyObjects()
```
mwaldt, line 165: This name is misleading now, did you mean to make 2 test functions?

Justintime50:
This implementation actually re-breaks `tax_identifiers`, continuing to test.

lib/EasyPost/Requestor.php
```diff
@@ -46,30 +48,38 @@ public static function utf8($value)
...
-            foreach ($d as $k => $v) {
-                $res[$k] = self::_encodeObjects($v);
+        } elseif (is_array($data)) {
+            $resource = array();
+            foreach ($data as $k => $v) {
+                $resource[$k] = self::_encodeObjects($v);
             }
 
-            return $res;
+            return array_filter($resource);
```
Justintime50, line 74: # Reviewer Note

The magic is in this function, we'll filter empty and null values here, this is tested at a root and nested level in this PR.

jontsai (approved):
LGTM

lib/EasyPost/Requestor.php
```diff
@@ -46,30 +48,38 @@ public static function utf8($value)
...
      */
-    private static function _encodeObjects($d)
+    private static function _encodeObjects($data)
     {
-        if ($d instanceof EasypostResource) {
-            return array("id" => self::utf8($d->id));
-        } elseif ($d === true) {
+        if (!$data) {
+            $data = array();
+        }
```
fiendish, line 60: This returns [] for false, but then below false should become 'false'.

PR #118, opened 2021-10-18
fix: corrects JSON encoding for various data types

### LINKED ISSUE #117: Error when trying to buy USPS postage with version 4.0.1
After upgrading to PHP SDK version 4.0.1, we are getting following error when any shipper tries to purchase USPS postage. Other carriers have not been affected. I think this might have to do with the switch to sending JSON payloads in version 4.0+. Can you please patch this bug asap? The latest PHP SDK versions are unusable in their current state. Thanks.

> Error parsing input.

Justintime50:
### Summary

This is an attempt at fixing JSON encoding for various data types. I tried to keep this as least destructive as possible (we could/should be using the `JsonSerializable` interface; however, I didn't want to introduce another larger rewrite if we could for now correct this with a simple pass.

This is accomplished now by:
1. Not using the `array_filter` which had some interesting side effects
2. Making any element that we previously would set as an empty array if no data was present to `null` instead
3. Sending all data as strings (as this was the previous behavior and what almost every field of our API expects)

### Tests

* null tests
* empty string tests
* string tests
* array tests (empty and filled)
* float tests
* integer tests
* boolean tests

I added a few tests to check encoding and updated others. Looking at the output of cassettes, I'm much happier with this approach than the previous as it's behaving much closer to what I'd expect.

Docs about type comparisons: https://www.php.net/manual/en/types.comparisons.php

Closes #117


lib/EasyPost/Requestor.php
```diff
@@ -65,13 +65,19 @@ private static function _encodeObjects($data)
             return 'true';
         } elseif ($data === false) {
             return 'false';
+        } elseif (is_integer($data)) {
```
jontsai, line 68: How about `is_numeric` instead?

https://www.php.net/manual/en/function.is-numeric.php

lib/EasyPost/Requestor.php
```diff
@@ -197,24 +203,18 @@ private function _curlRequest($method, $absUrl, $headers, $params, $myApiKey)
...
+            $absUrl = "$absUrl?$urlParams";
         } elseif ($method == 'post') {
             $curlOptions[CURLOPT_POST] = 1;
             $curlOptions[CURLOPT_POSTFIELDS] = json_encode($params);
         } elseif ($method == 'patch' || $method == 'put') {
             $curlOptions[CURLOPT_CUSTOMREQUEST] = strtoupper($method);
-            if (count($params) > 0) {
-                $curlOptions[CURLOPT_POSTFIELDS] = json_encode($params);
-            }
+            $curlOptions[CURLOPT_POSTFIELDS] = json_encode($params);
```
jontsai, line 213: I think the key observation to make is that `json_encode` happens here, and using the pure PHP built-in (https://www.php.net/manual/en/function.json-encode)

However, I believe, when reading #117, what we might want to do is to implement our own JSON encoder function that will do some pre-treatment to `$params` before passing it into the built-in `json_encode`

I was thinking that we could with your current approach of stringifying all numbers, or define a JSON mapping structure with types.

lib/EasyPost/Requestor.php
fiendish: There's already checks for true, false, and array and one that is intended to be for null but which is doing the wrong thing right now. What other types shouldn't all be strvaled?

lib/EasyPost/Requestor.php
fiendish: I'm not clear on why the param count checks were removed here.

fiendish:
> We are sending null instead of just dropping keys/objects if the user explicitly passes null. This is probably an acceptable behavior; however, we may want to simply drop them altogether?

It wouldn't hurt to throw a null key check inside the is_array foreach before recursing.

Like, if we're only looking for strval conversion here and not yet applying more explicit field type checks, maybe (keeping the param count > 0 checks later on instead of removing them)...
```
@@ -55,11 +55,9 @@
      */
     private static function _encodeObjects($data)
     {
-        if (!$data) {
-            $data = array();
-        }
-
-        if ($data instanceof EasypostResource) {
+        if (is_null($data)) {
+            return array();
+        } elseif ($data instanceof EasypostResource) {
             return array("id" => self::utf8($data->id));
         } elseif ($data === true) {
             return 'true';
@@ -68,12 +66,14 @@
         } elseif (is_array($data)) {
             $resource = array();
             foreach ($data as $k => $v) {
-                $resource[$k] = self::_encodeObjects($v);
+                if (!is_null($v)) {
+                    $resource[$k] = self::_encodeObjects($v);
+                }
             }
 
             return array_filter($resource);
         } else {
-            return self::utf8($data);
+            return self::utf8(strval($data));
         }
     }
```
Though I didn't test it

lib/EasyPost/Requestor.php
fiendish: Makes sense!

lib/EasyPost/Requestor.php
```diff
@@ -55,31 +55,31 @@ public static function utf8($value)
...
         } elseif ($data === true) {
             return 'true';
         } elseif ($data === false) {
             return 'false';
-        } elseif (is_integer($data)) {
-            return strval($data);
         } elseif (is_array($data)) {
             if (empty($data)) {
                 return null;
             }
```
fiendish: I just realized that this empty array null return kinda contradicts `is_null($data)` returning an empty array above. Like, if the caller passes null it gets back an empty array, and if the caller passes an empty array it gets back null. 🤕 

I think this wants to only happen for nested layers, in which case probably it should go next to `!is_null($v)` in the foreach?

fiendish (changes requested):
√

lib/EasyPost/Requestor.php
Justintime50: Ah, great call! This is evident in an empty `customs_items` null instead of an empty array. Fixed!

lib/EasyPost/Requestor.php
```diff
@@ -68,12 +68,16 @@ private static function _encodeObjects($data)
         } elseif (is_array($data)) {
             $resource = array();
             foreach ($data as $k => $v) {
-                $resource[$k] = self::_encodeObjects($v);
+                if (empty($data)) {
+                    return null;
+                } elseif (!is_null($v)) {
```
fiendish, line 73: Compound logic like this is a bit ugly, but I think in this form it would be more like
```suggestion
                if (!is_null($v)) and ($v !== "") and (!is_array($v) or !empty($v)) {
```

jontsai (approved):

lib/EasyPost/Requestor.php
```diff
@@ -197,24 +197,18 @@ private function _curlRequest($method, $absUrl, $headers, $params, $myApiKey)
         // Setup the HTTP method and params to use on the request
         if ($method == 'get') {
             $curlOptions[CURLOPT_HTTPGET] = 1;
-            if (count($params) > 0) {
-                $urlParams = self::_urlEncode($params);
-                $absUrl = "$absUrl?$urlParams";
-            }
+            $urlParams = self::_urlEncode($params);
+            $absUrl = "$absUrl?$urlParams";
```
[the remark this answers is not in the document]
fiendish: Agreed

jontsai (approved):

PR #119, opened 2021-10-20
chore: prepare v4.0.2 for release

Justintime50:
* Further fixes JSON encoding by dropping null key/values, sending values as strings where necessary (returning to previous behavior), and removing the `array_filter` from the previous release


mwaldt (approved):
number does go up

PR #120, opened 2021-11-01
chore: sets a default timeout of 60 seconds on requests

Justintime50:
This sets a default timeout of 30 seconds for connections and 60 seconds for requests. Currently, there is no timeout at all set and connections will stay open forever (or until the EasyPost servers kill them off).

60 seconds was chosen as it's long enough not to be hostile but short enough to not leave connections open for vast amounts of time.

Similarly, `CURLE_OPERATION_TIMEOUTED` is aliased to `CURLE_OPERATION_TIMEDOUT` which is now the proper way of writing this option. See https://www.php.net/manual/en/curl.constants.php for details.

This was end to end tested by setting the timeout to `1ms` and receiving a timeout error, then removing the timeout call to use the new default which sent the request through under 60 seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ### NEXT RELEASE
 
-* Set a default timeout of 60 seconds for connections and requests
+* Set a default timeout of 30 seconds for connections and 60 seconds for requests
 
 ### 4.0.2 2021-10-20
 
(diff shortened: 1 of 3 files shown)


fiendish:
> 60 seconds for connections

Is it 60 or 30?

Justintime50:
@fiendish sorry, forgot to update the description. Fixed.

fiendish (approved):

fiendish:
By the way, a lot of set/get function comments are swapped.

ajnassar (approved):
LGTM

PR #123, opened 2021-12-02
chore: removes the remainder of API keys from examples

Justintime50:
Removes the remainder of the hard-coded API keys from example files.


jchen293 (approved):
LGTM, only the hard-coded API key is changed.

nwithan8 (approved):
Yikes that these were there in the first place. LGTM

PR #125, opened 2022-01-12
fix: tracking create_list method and added tracking tests

Justintime50:
Adds tracker tests.

This exposed that the `Tracker::create_list` function broke when we switched to JSON encoded bodies. A small fix for this bug is included in this PR.

There is probably a better way of doing this, but it would require us to extend our classes with `JsonSerializable` but that would be a decent rewrite. I'd like to move on this as a "bandaid" fix for now and we can look into a longer term fix down the road.


nwithan8 (dismissed):
LGFM

nwithan8 (approved):
LGFM part 2.

jchen293 (approved):
lgtm

PR #128, opened 2022-01-24
feat: adds the update_brand method to the user object

Justintime50:
* Adds the `update_brand()` method on the user object

I was able to test this via a unit test which passed (eg: `$user->update_brand(array('color' => '#ffffff'))`; however, I did not leave the test nor check-in the cassette as it exposes user data (brand object, user object, child users, etc) and is not a safe cassette to keep in the repo. There are ways to sanitize the cassettes; however, after various attempts it would not work as we'd need to properly sanitize the data here. We'll make separate efforts to circle back to "safe user unit testing" across our libraries.


nwithan8 (approved):

jchen293 (approved):
Nice work, I tested it locally everything looks good to me!

PR #129, opened 2022-01-24
fix: adds ReturnTypeWillChange to temporarily ignore deprecation warnings

### LINKED ISSUE #124: PHP v8.1 incompatible
When called under PHP v8.1 there are several functions in EasyPostObject.php (ArrayAccess and Iterator methods) which need to be updated or add the comment #[\ReturnTypeWillChange] to prevent deprecation notice

Justintime50:
Adds `#[\ReturnTypeWillChange]` to some EasyPostObject methods to remove the deprecation warnings on PHP 8.1. The return type definitions will be required for PHP 9; however, because we will still be supporting PHP 7.3 for some time and the `mixed` return type which is required to fix this but not included until PHP 8 isn't possible at the moment, we'll temporarily ignore these deprecation notices until we drop support for PHP 7 altogether (I'd assume about the same time we adopt PHP 9 once it's out.)

Closes #124

### DIFF
--- a/lib/EasyPost/EasyPostObject.php
+++ b/lib/EasyPost/EasyPostObject.php
@@ -258,6 +258,7 @@ class EasyPostObject implements \ArrayAccess, \Iterator
      * @param string $k
      * @return mixed
      */
+    #[\ReturnTypeWillChange]
     public function offsetGet($k)
     {
         return array_key_exists($k, $this->_values) ? $this->_values[$k] : null;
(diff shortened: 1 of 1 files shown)


jchen293:
I have PHP 8.1.1 and did some testing on this, it looks like **offsetGet()** and **key()** are also throwing deprecation warnings.

Justintime50:
Great catch @jchen293! Must have been removed when I switched from return types to ignoring them for now. I've added the ignore comments to those two as well.

jchen293 (approved):

nwithan8 (approved):

PR #130, opened 2022-01-31
chore: documents all class properties (closes #96)

### LINKED ISSUE #96: Add @property documentation to response objects
First of all, thanks for the library.

It would be nice if you can add the expected properties into the response objects. This would help reduce typos for users of the library and allow IDEs to check for properties accessed using the magic get method.

Currently, I get warnings in my IDE when I'm trying to parse the response, because magic methods is not something myself, or the IDEs are fans of:
Screenshot from 2020-11-16 10-44-39

Adding property tags to the class documentation is a quick solution for that. Modifying the Rate class:
```php
<?php

namespace EasyPost;

/**
 * Class Rate
 * @package EasyPost
 * @property string $rate
 * @property string $service
 * @property string $carrier
 */
class Rate extends EasypostResource
{
    ...
}
```

results in no warnings and autocomplete:
Screenshot from 2020-11-16 10-46-34

Justintime50:
Documents all class properties per the request on #96.

Also went through each function's docstring to ensure it was correct, fixed grammar on most and added a few missing docblocks.

### DIFF
--- a/lib/EasyPost/Address.php
+++ b/lib/EasyPost/Address.php
@@ -2,10 +2,31 @@
 
 namespace EasyPost;
 
+/**
+ * @package EasyPost
+ * @property string $id
+ * @property string $object
+ * @property string $mode
+ * @property string $street1
+ * @property string $street2
+ * @property string $city
+ * @property string $state
+ * @property string $zip
+ * @property string $country
+ * @property bool $residential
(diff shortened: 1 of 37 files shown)


nwithan8 (approved):
LGTM, glad we finally got this done.

jchen293 (approved):
LGTM

PR #133, opened 2022-02-03
chore: use more modern array syntax

Justintime50:
PHP has two different ways of creating an array - `array()` or `[]`. This PR updates all occurrences of the older `array()` syntax to the newer and more widely accepted `[]`. This is also just now possible because we no longer need to retain the older `array` for backwards compatibility with PHP 5 as we just dropped support.


nwithan8 (approved):
LGFM, yay for modernization!

jchen293 (approved):
LGTM

PR #136, opened 2022-02-15
fix: test assertEquals order

Justintime50:
Swaps the order of all `assertEquals` tests to be `expected`, then `actual` like it should be.

Also found one `assertNotNull` that should have actually been an `assertEquals`.


jchen293 (approved):

nwithan8 (approved):
LGFM

PR #139, opened 2022-02-22
chore: assert against has_more key for pagination

Justintime50:
Adds `has_more` assertions to retrieving `all` records to ensure users can paginate properly.


jchen293 (approved):

nwithan8 (approved):
LGTM

PR #140, opened 2022-02-23
chore: use stronger test assertions where possible

Justintime50:
Uses stronger test assertions where possible, brought up by @jchen293 where these should have been written this way but were overlooked.


test/EasyPost/RefundTest.php
```diff
@@ -92,9 +90,9 @@ public function testRetrieve(object $refunds)
     {
         VCR::insertCassette('refunds/retrieve.yml');
 
-        $refund = Refund::retrieve($refunds['refunds'][0]);
+        $retrieved_refund = Refund::retrieve($refunds['refunds'][0]);
 
-        $this->assertInstanceOf('\EasyPost\Refund', $refund);
-        $this->assertStringMatchesFormat('rfnd_%s', $refund->id);
+        $this->assertInstanceOf('\EasyPost\Refund', $retrieved_refund);
+        $this->assertEquals($refunds['refunds'][0]->id, $retrieved_refund->id);
```
jchen293, line 96: We should able to compare objects for **testRetrieve()** in RefundTest. I was able to compare objects in Java.
```suggestion
        $this->assertEquals($refunds['refunds'][0], $retrieved_refund);
```

test/EasyPost/ReportTest.php
```diff
@@ -41,17 +41,17 @@ public function testCreatePaymentLogReport()
     {
         VCR::insertCassette('reports/createPaymentLogReport.yml');
 
-        $payment_log_report = Report::create([
-            "start_date" => Fixture::report_start_date(),
-            "end_date" => Fixture::report_end_date(),
-            "type" => "payment_log"
+        $report = Report::create([
```
jchen293, line 44: I would undo this change, so we can make the variable name more explicit.

test/EasyPost/ReportTest.php
```diff
@@ -129,107 +129,107 @@ public function testCreateTrackerReport()
...
     /**
      * Test retrieving a Payment Log report.
      *
-     * @param Report $payment_log_report
+     * @param Report $report
      * @return void
      * @depends testCreatePaymentLogReport
      */
-    public function testRetrievePaymentLogReport($payment_log_report)
+    public function testRetrievePaymentLogReport($report)
```
jchen293, line 152: Same here, some people might be confused on which **`report`** variable is this refers to.

jchen293 (changes requested):

test/EasyPost/ReportTest.php
Justintime50: This shouldn't matter because we are only testing one report per test, the type of the report is inherent. The name of report is also in the test name.

Having them more explicit would make sense if we had multiple reports in the same scope but we don't here.

test/EasyPost/ReportTest.php
Justintime50: I don't believe this is a problem because we specify `@depends testCreatePaymentLogReport` 2 lines above and the name of the test is `testRetrievePaymentLogReport`, they match up already.

test/EasyPost/RefundTest.php
Justintime50: I tried this one here but wasn't able to make it work due to how PHP's VCR works.

test/EasyPost/ReportTest.php
jchen293: I won't block it, but for readability, I prefer a variable name that is explicit.

jchen293 (dismissed):

nwithan8:
This might be the right ticket to bring this up on:
https://github.com/EasyPost/easypost-php/blob/master/test/EasyPost/CarrierAccountTest.php#L75
We should be comparing the retrieved carrier account and the original carrier account (or at least the IDs of each), a la https://github.com/EasyPost/easypost-php/blob/master/test/EasyPost/AddressTest.php#L90

Justintime50:
@nwithan8 indeed, I've corrected that.

nwithan8 (dismissed):

test/EasyPost/ScanFormTest.php
```diff
@@ -69,7 +69,7 @@ public function testRetrieve(ScanForm $scanform)
         $retrieved_scanform = ScanForm::retrieve($scanform->id);
 
         $this->assertInstanceOf('\EasyPost\ScanForm', $retrieved_scanform);
-        $this->assertStringMatchesFormat('sf_%s', $retrieved_scanform->id);
+        $this->assertStringMatchesFormat($scanform, $retrieved_scanform);
```
jchen293, line 72: The assertion here is wrong, it should be assertEquals()
```suggestion
        $this->assertEquals($scanform, $retrieved_scanform);
```

test/EasyPost/UserTest.php
```diff
@@ -49,6 +49,7 @@ public function testCreate()
 
         $this->assertInstanceOf('\EasyPost\User', $user);
         $this->assertStringMatchesFormat('user_%s', $user->id);
+        $this->assertEquals($user->name, 'Test User');
```
jchen293, line 52: Order is wrong here.
```suggestion
        $this->assertEquals('Test User', $user->name);
```

jchen293 (changes requested):

test/EasyPost/ScanFormTest.php
Justintime50: Nice catch.

jchen293 (approved):

nwithan8:
Caught another one: https://github.com/EasyPost/easypost-php/blob/master/test/EasyPost/InsuranceTest.php#L76

Justintime50:
@nwithan8 that should have been fixed in this PR already. The link you gave is for the master branch, not this PR.

nwithan8 (approved):

PR #143, opened 2022-03-09
fix batch test param.

jchen293:
fix batch param


Justintime50:
This isn't necessary as the current fixture used is already correct.

PR #145, opened 2022-03-22
[feat] Support for `columns` and `additional_columns` on Report creation

nwithan8:
This PR:
- Adds support for specifying `columns` and `additional_columns` (both lists of strings) in the parameters dictionary passed into the `Report->create()` function.

Unfortunately, since the API response does not indicate the columns used for a report, the only way to ensure the report actually includes the columns requested would be to make a live API call, download the generated CSV report and parse its column headers. That is a complex unit test with many points of failure and prevents us from using VCR.

Instead, we simply assume that if no error is thrown during the API call to generate and retrieve the report, that the report contains the correct columns. If it doesn't, that constitutes a server-side error that is out of our control.

### DIFF
--- a/lib/EasyPost/Report.php
+++ b/lib/EasyPost/Report.php
@@ -67,6 +67,34 @@ class Report extends EasypostResource
      */
     public static function create($params = null, $apiKey = null)
     {
+        if ((isset($params['columns']) && is_array($params['columns'])) || (isset($params['additional_columns']) && is_array($params['additional_columns']))) {
+            $columns = "";
+            if (isset($params['columns'])) {
+                $columns = $params['columns'];
+                unset($params['columns']);
+            }
+
+            $additional_columns = "";
+            if (isset($params['additional_columns'])) {
+                $additional_columns = $params['additional_columns'];
+                unset($params['additional_columns']);
+            }
+
+            $urlMod = "?";
+
+            if (is_array($columns)) {
+                foreach ($columns as $column) {
+                    $urlMod .= "columns[]=" . $column . "&";
+                }
+            }
+
+            if (is_array($additional_columns)) {
+                foreach ($additional_columns as $additional_column) {
+                    $urlMod .= "additional_columns[]=" . $additional_column . "&";
+                }
+            }
+        }
+
         if (!isset($params['type'])) {
             throw new Error('Undetermined Report Type');
         } else {
--- a/lib/EasyPost/Report.php
+++ b/lib/EasyPost/Report.php
@@ -67,47 +67,39 @@ class Report extends EasypostResource
      */
     public static function create($params = null, $apiKey = null)
     {
-        if ((isset($params['columns']) && is_array($params['columns'])) || (isset($params['additional_columns']) && is_array($params['additional_columns']))) {
-            $columns = "";
-            if (isset($params['columns'])) {
-                $columns = $params['columns'];
-                unset($params['columns']);
-            }
-
-            $additional_columns = "";
-            if (isset($params['additional_columns'])) {
-                $additional_columns = $params['additional_columns'];
-                unset($params['additional_columns']);
-            }
-
-            $urlMod = "?";
-
-            if (is_array($columns)) {
-                foreach ($columns as $column) {
-                    $urlMod .= "columns[]=" . $column . "&";
-                }
-            }
-
-            if (is_array($additional_columns)) {
-                foreach ($additional_columns as $additional_column) {
-                    $urlMod .= "additional_columns[]=" . $additional_column . "&";
-                }
-            }
-        }
-
         if (!isset($params['type'])) {
             throw new Error('Undetermined Report Type');
-        } else {
-            $type = $params['type'];
-
-            self::_validate($params, $apiKey);
-            $requestor = new Requestor($apiKey);
-
-            $url = self::reportUrl($type);
-
-            list($response, $apiKey) = $requestor->request('post', $url, $params, true);
-            return Util::convertToEasyPostObject($response, $apiKey);
         }
+
+        $url = self::reportUrl($params['type']);
+
+        $urlMod = "";
+
+        if ((isset($params['columns']) && is_array($params['columns']))) {
+            foreach ($params['columns'] as $column) {
+                $urlMod .= "columns[]=" . $column . "&";
+            }
+            // Removing from params since already used in query params
+            unset($params['columns']);
+        }
+
+        if ((isset($params['additional_columns']) && is_array($params['additional_columns']))) {
+            foreach ($params['additional_columns'] as $additional_column) {
+                $urlMod .= "additional_columns[]=" . $additional_column . "&";
+            }
+            // Removing from params since already used in query params
+            unset($params['additional_columns']);
+        }
+
+        if ($urlMod != "") {
+            $url .= "?" . $urlMod;
+        }
+
+        self::_validate($params, $apiKey);
+        $requestor = new Requestor($apiKey);
+
+        list($response, $apiKey) = $requestor->request('post', $url, $params, true);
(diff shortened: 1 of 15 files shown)


lib/EasyPost/Report.php
```diff
@@ -67,6 +67,34 @@ public static function all($params = null, $apiKey = null)
...
+
+            if (is_array($columns)) {
+                foreach ($columns as $column) {
+                    $urlMod .= "columns[]=" . $column . "&";
+                }
+            }
+
+            if (is_array($additional_columns)) {
+                foreach ($additional_columns as $additional_column) {
+                    $urlMod .= "additional_columns[]=" . $additional_column . "&";
```
jchen293, line 93: this `urlMod` is not being used anywhere, shouldn't this `urlMod replace the `url` when making the request call?

Justintime50: Jack is correct, this isn't getting used and should be `$url`, please re-record the cassettes to ensure this worked.

lib/EasyPost/Report.php
```diff
@@ -67,6 +67,34 @@ public static function all($params = null, $apiKey = null)
      */
     public static function create($params = null, $apiKey = null)
     {
+        if ((isset($params['columns']) && is_array($params['columns'])) || (isset($params['additional_columns']) && is_array($params['additional_columns']))) {
```
Justintime50, line 70: There is too much nesting here. These should be individual if statements and not using the `||` operator since they should be mutually exclusive. Two blocks of code, one for `columns` and one for `additional_columns`.

lib/EasyPost/Report.php
```diff
@@ -67,6 +67,34 @@ public static function all($params = null, $apiKey = null)
...
+            $columns = "";
+            if (isset($params['columns'])) {
```
Justintime50, line 72: Per my comment above, if we split these into separate blocks, there will be no need to run multiple if statements since it will no longer be nested.

test/EasyPost/ReportTest.php
```diff
@@ -142,6 +142,50 @@ public function testCreateTrackerReport()
...
+            'start_date' => Fixture::report_start_date(),
+            'end_date' => Fixture::report_end_date(),
+            'type' => 'shipment',
+            'columns' => ['usps_zone']
+        ]);
+
+        $this->assertInstanceOf('\EasyPost\Report', $report);
+        // Reports are queued, so we can't retrieve it immediately.
+        // Verifying columns would require parsing the CSV.
+        // Verify parameters sent correctly by checking the URL in the cassette.
```
Justintime50, line 164: Please move this above the assertion. Code that's sitting by itself without being attached to anything is confusing. Having it above will help explain why there's only a single assertion and prompt users to go elsewhere to find more data.

test/EasyPost/ReportTest.php
```diff
@@ -142,6 +142,50 @@ public function testCreateTrackerReport()
...
+            'start_date' => Fixture::report_start_date(),
+            'end_date' => Fixture::report_end_date(),
+            'type' => 'shipment',
+            'additional_columns' => ['from_name', 'from_company']
+        ]);
+
+        $this->assertInstanceOf('\EasyPost\Report', $report);
+        // Reports are queued, so we can't retrieve it immediately.
+        // Verifying columns would require parsing the CSV.
+        // Verify parameters sent correctly by checking the URL in the cassette.
```
Justintime50, line 186: Comment above.

Justintime50 (changes requested):

test/EasyPost/ReportTest.php
nwithan8: Resolved in e97ee36e508d4a6395d03c8575079749f3aec069

test/EasyPost/ReportTest.php
nwithan8: Resolved in e97ee36e508d4a6395d03c8575079749f3aec069

lib/EasyPost/Report.php
nwithan8: Resolved in e97ee36e508d4a6395d03c8575079749f3aec069

lib/EasyPost/Report.php
nwithan8: Was modeling after our Address verification setup.
Resolved in e97ee36e508d4a6395d03c8575079749f3aec069

lib/EasyPost/Report.php
nwithan8: Resolved in e97ee36e508d4a6395d03c8575079749f3aec069

lib/EasyPost/Report.php
```diff
@@ -69,17 +69,38 @@ public static function create($params = null, $apiKey = null)
     {
         if (!isset($params['type'])) {
             throw new Error('Undetermined Report Type');
-        } else {
-            $type = $params['type'];
+        }
 
-            self::_validate($params, $apiKey);
-            $requestor = new Requestor($apiKey);
+        $type = $params['type'];
```
Justintime50, line 74: This doesn't really help since it's only being used once, there is no need for the intermediate variable and instead could just be used inline in the `$url` var below OR this could come above line ~70 and reused in that if check which would make sense to have a var since it would then be used twice.

lib/EasyPost/Report.php
```diff
@@ -69,17 +69,38 @@ public static function create($params = null, $apiKey = null)
...
 
-            list($response, $apiKey) = $requestor->request('post', $url, $params, true);
-            return Util::convertToEasyPostObject($response, $apiKey);
+        $urlMod = "";
+
+        if ((isset($params['columns']) && is_array($params['columns']))) {
+            foreach ($params['columns'] as $column) {
+                $urlMod .= "columns[]=" . $column . "&";
+            }
+            unset($params['columns']);
```
Justintime50, line 84: Please add a comment as to why we are unsetting these here

lib/EasyPost/Report.php
```diff
@@ -69,17 +69,38 @@ public static function create($params = null, $apiKey = null)
...
+            foreach ($params['additional_columns'] as $additional_column) {
+                $urlMod .= "additional_columns[]=" . $additional_column . "&";
+            }
+            unset($params['additional_columns']);
         }
+
+        if ($urlMod != "") {
+            $url = substr($url, 0, -1);
+            $url .= "?" . $urlMod;
+        }
```
Justintime50, line 97: Why do this instead of just appending to the already existent `$url`?

Justintime50 (changes requested):

lib/EasyPost/Report.php
nwithan8: Addressed in 6d220378d34ba8ed3dccaab89bb5493dddd7a2c4

lib/EasyPost/Report.php
```diff
@@ -69,17 +69,38 @@ public static function create($params = null, $apiKey = null)
...
+                $urlMod .= "columns[]=" . $column . "&";
+            }
+            unset($params['columns']);
+        }
+
+        if ((isset($params['additional_columns']) && is_array($params['additional_columns']))) {
+            foreach ($params['additional_columns'] as $additional_column) {
+                $urlMod .= "additional_columns[]=" . $additional_column . "&";
+            }
+            unset($params['additional_columns']);
```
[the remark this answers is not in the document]
nwithan8: 6d220378d34ba8ed3dccaab89bb5493dddd7a2c4

lib/EasyPost/Report.php
nwithan8: 6d220378d34ba8ed3dccaab89bb5493dddd7a2c4

lib/EasyPost/Report.php
nwithan8: Addressed in 6d220378d34ba8ed3dccaab89bb5493dddd7a2c4

PR #154, opened 2022-04-08
tests: more fixtures, dynamic dates, better instructions

Justintime50:
The goal of this PR is to make the test suite much more bullet proof without the need to manually alter fixture data in code as well as making it easier for 3rd party contributors to test this library.

- Cleaned up various tests
- Provided more fixtures to DRY up tests
- Use dynamic date generation instead of hard coded dates
- Removed the hard-coded child user ID
- Added helpful instructions to the README on how to test this library (from a 3rd party contributor perspective)


test/EasyPost/ReportTest.php
```diff
@@ -33,62 +33,18 @@ public static function tearDownAfterClass(): void
...
-        // Return so other tests can reuse this object
-        return $report;
-    }
-
-    /**
-     * Test creating a Refund report.
-     *
-     * @return Report
-     */
-    public function testCreateRefundReport()
```
nwithan8, line 62: We don't want to do tests for each type of report anymore?

Justintime50: Nah, this was me being a bit excessive on our first pass. The only difference with these tests is passing in a different `type`. None of the libs have any logic related to differing report types and is all handled on the API end so having 12 tests instead of 2 when there is nothing new to assert against just adds extra bloat to the project.

Justintime50: AKA: We don't gain anything from the extra 10 tests.

jchen293 (approved):

nwithan8 (approved):

PR #157, opened 2022-04-13
remove additional code that add params to URL query for post method.

jchen293:
# Pull Request Type

# Description

This PR removes code that adds `strict` and `verify_strict` to the URL query when the API method is POST.  Small update the fixture and adjust the unit test.

# Testing

re-record existing cassettes and all unit tests passed.

### DIFF
--- a/lib/EasyPost/Address.php
+++ b/lib/EasyPost/Address.php
@@ -58,43 +58,23 @@ class Address extends EasypostResource
      */
     public static function create($params = null, $apiKey = null)
     {
-        $urlMod = "";
+        $wrapped_params = [];
 
-        if ((isset($params['verify']) && is_array($params['verify'])) || (isset($params['verify_strict']) && is_array($params['verify_strict']))) {
-            $verify = "";
-            if (isset($params['verify'])) {
-                $verify = $params['verify'];
-                unset($params['verify']);
-            }
-
-            $verify_strict = "";
-            if (isset($params['verify_strict'])) {
-                $verify_strict = $params['verify_strict'];
-                unset($params['verify_strict']);
-            }
-
-            $urlMod = "?";
-
-            if (is_array($verify)) {
-                foreach ($verify as $verification) {
-                    $urlMod .= "verify[]=" . $verification . "&";
-                }
-            }
-
-            if (is_array($verify_strict)) {
-                foreach ($verify_strict as $verification_strict) {
-                    $urlMod .= "verify_strict[]=" . $verification_strict . "&";
-                }
-            }
+        if (isset($params['verify']) && is_array($params['verify'])) {
+            $verify = $params['verify'];
+            unset($params['verify']);
+            $wrapped_params['verify'] = $verify;
         }
 
-        if (!isset($params['address']) || !is_array($params['address'])) {
-            $clone = $params;
-            unset($params);
-            $params['address'] = $clone;
+        if (isset($params['verify_strict']) && is_array($params['verify_strict'])) {
+            $verify_strict = $params['verify_strict'];
+            unset($params['verify_strict']);
+            $wrapped_params['verify_strict'] = $verify_strict;
         }
 
-        return self::_create(get_class(), $params, $apiKey, $urlMod);
+        $wrapped_params["address"] = $params;
+
+        return self::_create(get_class(), $wrapped_params, $apiKey);
     }
 
     /**
@@ -140,10 +120,6 @@ class Address extends EasypostResource
 
         if (isset($response['address'])) {
             $verified_address = Util::convertToEasyPostObject($response['address'], $apiKey);
-            if (!empty($response['message'])) {
-                $verified_address->message = $response['message'];
-                $verified_address->_immutableValues[] = 'message';
-            }
 
             return $verified_address;
         } else {
--- a/test/cassettes/shipments/createWithIds.yml
+++ b/test/cassettes/shipments/createWithIds.yml
@@ -24,60 +24,59 @@
             x-download-options: noopen
             x-permitted-cross-domain-policies: none
             referrer-policy: strict-origin-when-cross-origin
-            x-ep-request-uuid: 04d9384d622b9c76e787ccdc002c4ab4
+            x-ep-request-uuid: b8f5271662586ddfe799f1df004b1fb3
             cache-control: 'no-cache, no-store'
             pragma: no-cache
             expires: '0'
-            location: /api/v2/addresses/adr_9da81c01a16d11ec9e86ac1f6b0a0d1e
+            location: /api/v2/addresses/adr_4c94cdebbc2411ec8adeac1f6b0a0d1e
             content-type: 'application/json; charset=utf-8'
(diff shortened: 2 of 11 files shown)


test/cassettes/shipments/createWithIds.yml
```diff
@@ -103,61 +102,59 @@
             x-download-options: noopen
             x-permitted-cross-domain-policies: none
             referrer-policy: strict-origin-when-cross-origin
-            x-ep-request-uuid: 04d9384a622b9c77e787ccdd002c4ad6
+            x-ep-request-uuid: f99c9a8862574186ff11cb0000520177
             cache-control: 'no-cache, no-store'
             pragma: no-cache
             expires: '0'
-            location: /api/v2/parcels.prcl_3274f0f657194a8893c6e5da7aa8cd5e
```
Justintime50, line 110: Hmm... this should be a `/` not a `.`, can we look into this?

lib/EasyPost/Address.php
```diff
@@ -58,43 +58,7 @@ public static function all($params = null, $apiKey = null)
...
-                    $urlMod .= "verify_strict[]=" . $verification_strict . "&";
-                }
-            }
-        }
-
-        if (!isset($params['address']) || !is_array($params['address'])) {
-            $clone = $params;
-            unset($params);
-            $params['address'] = $clone;
-        }
```
Justintime50, line 95: This still needs to be here. Basically it's a convenience for the user if they don't wrap their params in `address`, we'll do it for them.

Justintime50 (changes requested):

lib/EasyPost/Address.php
jchen293: I just removed it from Java lib, the API endpoint still takes the param if it's not wrapped in `address`.

Justintime50: Yeah we definitely need this.

Additionally we need to ensure the `verify` and `verify_strict` params are at the top level outside of `address`.

test/cassettes/shipments/createWithIds.yml
Justintime50: I briefly looked at this cursed problem... can't determine why it's showing up like this but it's working as intended.

Justintime50:
Please also remove the following block of code in the `verify` function that accidentally got skipped (since it's related):

```php
            if (!empty($response['message'])) {
                $verified_address->message = $response['message'];
                $verified_address->_immutableValues[] = 'message';
            }
```

lib/EasyPost/Address.php
jchen293: I have added this back

lib/EasyPost/Address.php
```diff
@@ -58,43 +58,25 @@ public static function all($params = null, $apiKey = null)
...
-
         if (!isset($params['address']) || !is_array($params['address'])) {
             $clone = $params;
             unset($params);
             $params['address'] = $clone;
         }
 
-        return self::_create(get_class(), $params, $apiKey, $urlMod);
+        if ((isset($params['address']['verify']) && is_array($params['address']['verify']))) {
```
Justintime50, line 67: This should be checking for the `verify` and `verify_strict` params at the root level and not nested in `address`.

Justintime50 (changes requested):

lib/EasyPost/Address.php
jchen293: What I have should be correct. I am checking `verify` and `verify_strict` in `address` which is nested, see lines 61 - 64.

Justintime50: Not quite.

Lines 61-65 above are a convenience to wrap the parameters in `address` **only if the user didn't already do this**. What we should be doing is checking if the `params` have a direct key of `verify` and `verify_strict` as it should not be nested inside of an `address` key inside of params. We want the check to be up one level. For instance, in the Java/Ruby PRs you just put up which are done correctly, you aren't checking for `verify` inside of `address`, you are looking for it at the root of `params` which is the change we need to also make here.

lib/EasyPost/Address.php
```diff
@@ -140,10 +122,6 @@ public function verify()
 
         if (isset($response['address'])) {
             $verified_address = Util::convertToEasyPostObject($response['address'], $apiKey);
-            if (!empty($response['message'])) {
```
Justintime50, line 143: Thanks!

Justintime50 (dismissed):

lib/EasyPost/Address.php
jchen293: I think either way should work, Java/Ruby checks `verify` and `verify_strict` before the wrapper but PHP is after. I will change the code to stay consistent.

Justintime50: It "works" now because of the way we are passing the data in; however, this will break for some customers. We should take the most generic/safest route. If you look at how the code used to work before these changes, you'll notice that they are doing what I'm suggesting: `$params['verify']`, etc.

jchen293: There is `unset` which removes the `verify` and `verify_strict` inside of `address`, `verify` and `verify_strict` will be added in the top-level param.

nwithan8 (approved):
lgtm

lib/EasyPost/Address.php
```diff
@@ -58,43 +58,23 @@ public static function all($params = null, $apiKey = null)
...
-                    $urlMod .= "verify[]=" . $verification . "&";
-                }
-            }
-
-            if (is_array($verify_strict)) {
-                foreach ($verify_strict as $verification_strict) {
-                    $urlMod .= "verify_strict[]=" . $verification_strict . "&";
-                }
-            }
+        if ((isset($params['verify']) && is_array($params['verify']))) {
```
Justintime50, line 63:
```suggestion
        if (isset($params['verify'] && is_array($params['verify'])) {
```

lib/EasyPost/Address.php
```diff
@@ -58,43 +58,23 @@ public static function all($params = null, $apiKey = null)
...
+            $verify = $params['verify'];
+            unset($params['verify']);
+            $wrapped_params['verify'] = $verify;
         }
 
-        if (!isset($params['address']) || !is_array($params['address'])) {
-            $clone = $params;
-            unset($params);
-            $params['address'] = $clone;
+        if ((isset($params['verify_strict']) && is_array($params['verify_strict']))) {
```
Justintime50, line 69:
```suggestion
        if (isset($params['verify_strict']) && is_array($params['verify_strict'])) {
```

Justintime50 (changes requested):
We shouldn't need these extra brackets.

lib/EasyPost/Address.php
jchen293: resolved

lib/EasyPost/Address.php
jchen293: resolved

Justintime50 (approved):

nwithan8 (approved):

PR #167, opened 2022-05-12
- Adds the EndShipper beta class

jchen293:
# Description

Adds the new EndShipper beta class with `create`, `retrieve`, `all`, and `save` functions. This is the fully-qualified address object that the USPS will start requiring shortly.

# Testing

Add unit tests

# Pull Request Type


Justintime50 (approved):

lib/EasyPost/Beta/EndShipper.php
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Update (save) an EndShipper object.
+     *
+     * @return $this
+     */
+    public function save()
+    {
+        // We are passing the `Address` class here so that the request gets properly wrapped in the required object.
+        return $this->_save('EasyPost\Address', true);
```
nwithan8, line 70: Are we still using Address, or are we making a standalone EndShipper object, per our internal conversation today?

nwithan8 (changes requested):

lib/EasyPost/Beta/EndShipper.php
jchen293: This is something else that has to do with the API endpoint. The `EndShipper update endpoint` accepts `address` in the parameter instead of `end_shipper`. Nothing has to do with the conversation today.

Justintime50: Ha yeah, we dynamically determine the endpoint to hit based on the class name. All others pass in the `get_class()` function but here we need to specify the Address class so that it properly wraps all the params in `address` instead of `end_shipper` per the docs.

nwithan8 (approved):

test/EasyPost/Beta/EndShipperTest.php
```diff
@@ -0,0 +1,113 @@
...
+        $end_shipper->name = 'Jack Sparrow';
+        $end_shipper->company = 'EasyPost';
+        $end_shipper->street1 = '388 Townsend St';
+        $end_shipper->street2 = 'Apt 20';
+        $end_shipper->city = 'San Francisco';
+        $end_shipper->state = 'CA';
+        $end_shipper->zip = '94107';
+        $end_shipper->country = 'US';
+        $end_shipper->phone = '9999999999';
+        $end_shipper->email = 'test@example.com';
```
Justintime50, line 106: Nit: We could have reassigned these to the individual fixture values for each field so we didn't need to reuse these values.

Justintime50 (approved):

PR #174, opened 2022-05-28
Fix default value for `$beta` parameter in `refresh()`

nwithan8:
# Description

- Set default value for $beta parameter in refresh() function
- Update Batches example for $beta parameter

# Testing

All tests pass, no need to re-record

NOTE: `refresh()` function not covered by test suite. Suggestions for unit test?

# Pull Request Type


lib/EasyPost/EasypostResource.php
```diff
@@ -68,7 +68,7 @@ public function instanceUrl()
      * @return $this
      * @throws \EasyPost\Error
      */
-    public function refresh($beta)
+    public function refresh($beta = false): EasypostResource
```
Justintime50, line 71: I'd prefer only adjusting this line (adding the default value of `false` for this whole PR and remove passing in false every time this function is called. Setting a parameter to its default is redundant and would read better by only passing when overriding (such as the case of saying a certain refresh function is beta or `true`). Thoughts?

We are going to revisit the return types as they will be required in PHP 9 but many of the return types aren't yet supported at runtime in PHP 7.x. Let's not add this just yet for that reason (it'll yell at ya)

Justintime50 (commented):
We need to call this out in the CHANGELOG.

As for not testing this, we actually do as it gets called for every `retrieve` function; however, it didn't fail there because we were passing in the default `beta` value of the retrieve function which was false. In the case of the user reporting this, I think the only way that happened was because they called `refresh()` themselves. I don't know that this function should even be public, it's really a helper/wrapper for our actual public functions. My suggestion would be to deprecate this now (with warning if directly accessed, if nothing else a comment in code) and create a task to make this private in the next major release. If someone wanted to refresh their object, they really need to be retrieving it (which then calls this refresh function).

jchen293 (approved):

lib/EasyPost/EasypostResource.php
```diff
@@ -65,11 +65,14 @@ public function instanceUrl()
     /**
      * Refresh the object from the API.
      *
+     * @deprecated This function will be removed in the next major version.
```
Justintime50, line 68: I don't want to lead to confusion.
```suggestion
     * @deprecated This function will be made unavailable in the next major version.
```

Justintime50 (dismissed):
Code is great.

Prior to merging this, I want to know how many of our other libs have a publicly available `refresh` function. There's actually an already "internal" `refreshFrom` function that gets called when a user calls `refresh`. I'd like to determine a holistic "refresh" approach across libs before we move to deprecating something.

Alternatively, we could not mark it for deprecation (yet), file an internal task to look into "refreshing objects" across libs, simply fix the beta default param here for now and cut a new release, and circle back if we wanted to get this out ASAP to fix this bug. I'd probably suggest this course of action for a speedy bug fix which will allow us more time to come up with the best long-term solution about refreshing.

Justintime50 (approved):
Thanks. I know cutting this way back isn't super fun, but I'm a big fan of taking a bit more time on something to do it right once then needing to come back and tweak it later.

jchen293 (approved):

PR #177, opened 2022-05-31
Add Billing functions.

jchen293:
# Description

Adds PaymentMethod.all(), CreditCard.fund(), and CreditCard.delete() functions

# Testing

Added unit test for retrieving all payment methods
Can't test `CreditCard.fund` and `CreditCard.delete` due to the lack of an available real credit card in the tests

# Pull Request Type


nwithan8 (dismissed):

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## Next Release
+
+- Adds `PaymentMethod.all()`, `CreditCard.fund()`, and `CreditCard.delete()` functions
```
Justintime50, line 5: This is not the proper usage for PHP. Be sure to mirror what the tests use to be sure.

```suggestion
- Adds `PaymentMethod::all()`, `CreditCard::fund()`, and `CreditCard::delete()` functions
```

lib/EasyPost/CreditCard.php
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public static function fund($amount, $primary_or_secondary = 'primary', $api_key = null)
+    {
+        $payment_methods = PaymentMethod::all();
+        $payment_method_map = [
+            'primary' => 'primary_payment_method',
+            'secondary' => 'secondary_payment_method'
+        ];
+        $payment_method_to_use = $payment_method_map[$primary_or_secondary] ?? null;
+        $card_id = $payment_methods->$payment_method_to_use->id;
```
Justintime50, line 33: Have you tested this to ensure a runtime error isn't thrown if it can't lookup the key?

lib/EasyPost/CreditCard.php
```diff
@@ -0,0 +1,60 @@
...
+
+        if ($payment_method_to_use === null || $card_id === null || strpos($card_id, 'card_') !== 0) {
+            throw new Error('The chosen payment method is not a credit card. Please try again.');
+        }
+
+        $url = self::classUrl(get_class()) . '/' . $card_id . '/charges';
```
Justintime50, line 39: We can use string interpolation here. The only thing that can't be in the string is the function.

```suggestion
        $url = self::classUrl(get_class()) . "/$card_id/charges";
```

lib/EasyPost/CreditCard.php
```diff
@@ -0,0 +1,60 @@
...
+    /**
+     * Delete a credit card.
+     *
+     * @param string $credit_card_id
+     * @param string $api_key
+     * @return mixed
+     */
+    public static function delete($credit_card_id, $api_key = null)
+    {
+        $url = self::classUrl(get_class()) . '/' . $credit_card_id;
```
Justintime50, line 55:
```suggestion
        $url = self::classUrl(get_class()) . "/$credit_card_id";
```

Justintime50 (changes requested):

lib/EasyPost/CreditCard.php
jchen293: Yes. `??` is going to assign the value to null if the key does not exist. here is the link for testing: https://onecompiler.com/php/3y5ta6kyu

Justintime50: Right. I mean line 33. If you try to assign the card Id to `$payment_methods->null->id`, it'll blow up because you can't get an ID from null. You need to also have a check on `card_id` to be null if `payment_method_to_use` is also null.

jchen293: Got it, I just tested it, it will throw an error in `$payment_methods->null->id`, I will add a conditional statement to avoid it.

Justintime50 (approved):

nwithan8 (approved):

PR #188, opened 2022-06-29
Combine CreditCard and BankAccount fund() and delete() into one method.

jchen293:
# Description

Combine credit card and bank account `fund()` and `delete()` into one generic method.

# Testing

Can't perform the test due to lack of real payment methods

# Pull Request Type


lib/EasyPost/Billing.php
```diff
@@ -12,10 +12,10 @@
  * @property string $expiration_year
  * @property string $brand
  */
-class CreditCard extends EasypostResource
+class Billing extends EasypostResource
 {
     /**
-     * Fund a credit card.
+     * Fund your EasyPost wallet by charging your primary or secondary card on file.
```
Justintime50, line 18:
```suggestion
     * Fund your EasyPost wallet by charging your primary or secondary payment method.
```

lib/EasyPost/Billing.php
```diff
@@ -53,11 +40,37 @@ public static function fund($amount, $primary_or_secondary = 'primary', $api_key
      * @param string $api_key
      * @return mixed
      */
-    public static function delete($credit_card_id, $api_key = null)
+    public static function delete($primary_or_secondary, $api_key = null)
```
Justintime50, line 43: Issue: This function should be on the `PaymentMethod` class. You can't delete "a billing", but you can delete "a payment method".

lib/EasyPost/Billing.php
```diff
@@ -24,22 +24,9 @@ class CreditCard extends EasypostResource
...
-            $card_id = $payment_methods->$payment_method_to_use->id;
-        }
-
-        if ($payment_method_to_use === null || $card_id === null || strpos($card_id, 'card_') !== 0) {
-            throw new Error('The chosen payment method is not a credit card. Please try again.');
-        }
+        $payment_method_info = Billing::get_payment_info($primary_or_secondary);
 
-        $url = self::classUrl(get_class()) . "/$card_id/charges";
+        $url = $payment_method_info[0] . "/$payment_method_info[1]/charges";
```
Justintime50, line 29: Issue: We should be unpacking these values or assigning them to variables. Whenever you pass a variabled index to something, it will lead to confusion because we don't actually know what these values are.

lib/EasyPost/Billing.php
```diff
@@ -53,11 +40,37 @@ public static function fund($amount, $primary_or_secondary = 'primary', $api_key
...
-        $url = self::classUrl(get_class()) . "/$credit_card_id";
+        $payment_method_info = Billing::get_payment_info($primary_or_secondary);
+
+        $url = $payment_method_info[0] . "/$payment_method_info[1]";
         $requestor = new Requestor($api_key);
         list($response, $api_key) = $requestor->request('delete', $url);
         return Util::convertToEasyPostObject($response, $api_key);
     }
+
+    public static function get_payment_info($primary_or_secondary = 'primary')
```
Justintime50, line 53: Issue: This is an internal function that our public functions use under the hood and as such, it should not be publicly exposed.

```suggestion
    private static function get_payment_info($primary_or_secondary = 'primary')
```

Justintime50 (changes requested):

lib/EasyPost/Billing.php
nwithan8: Or, alternatively, keep it here under billing, and rename the function. Something like `EasyPost.Billing.deletePaymentMethod()`

jchen293: Our plan is to move the functions to PaymentMethod class and remove the Billing class since it makes more sense for `PaymentMethod::all()` instead of `Billing::All()`

lib/EasyPost/Billing.php
jchen293: resolved

lib/EasyPost/Billing.php
jchen293: good suggestion, resolved

Justintime50 (changes requested):
Per our conversations as a team, let's use the following:

- `Billing::retrieve_payment_methods()`
- `Billing::fund_wallet(2000, 'primary')`
- `Billing::delete_payment_method('primary')`

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+     * @return mixed
+     */
+    public static function retrieve_payment_methods($params = null, $apiKey = null)
+    {
+        $payment_methods = self::_all("paymentMethod", $params, $apiKey);
+
+        if ($payment_methods->id == null) {
+            throw new Error('Billing has not been setup for this user. Please add a payment method.');
+        }
+        return $payment_methods;
```
Justintime50, line 28: Nit: return statements should be separated by a newline

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+    {
+        $payment_method_info = Billing::get_payment_info($primary_or_secondary);
+        $payment_method_endpoint = $payment_method_info[0];
+        $payment_method_id = $payment_method_info[1];
+
+        $url = $payment_method_endpoint . "/$payment_method_id/charges";
+        $wrapped_params = ['amount' => $amount];
+        $requestor = new Requestor($api_key);
+        list($response, $api_key) = $requestor->request('post', $url, $wrapped_params);
+        return Util::convertToEasyPostObject($response, $api_key);
```
Justintime50, line 49: Nit: return statements should be separated by a newline

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+        ];
+        $payment_method_to_use = $payment_method_map[$primary_or_secondary] ?? null;
+
+        if ($payment_methods->$payment_method_to_use != null) {
+            $payment_method_id = $payment_methods->$payment_method_to_use->id;
+        }
+
+        if ($payment_method_to_use !== null && $payment_method_id !== null) {
+            if (strpos($payment_method_id, 'card_') !== 0) {
+                return array('credit_cards', $payment_methods->$payment_method_to_use->id);
```
Justintime50, line 92:
```suggestion
                return ['credit_cards', $payment_methods->$payment_method_to_use->id];
```

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+            } else if (strpos($payment_method_id, 'bank_') !== 0) {
+                return array('bank_accounts', $payment_methods->$payment_method_to_use->id);
```
Justintime50, line 94:
```suggestion
                return ['bank_accounts', $payment_methods->$payment_method_to_use->id];
```

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+     *
+     * @param string $primary_or_secondary
+     * @param string $api_key
+     * @return mixed
+     */
+    public static function delete_payment_method($primary_or_secondary, $api_key = null)
+    {
+        $payment_method_info = Billing::get_payment_info($primary_or_secondary);
+        $payment_method_endpoint = $payment_method_info[0];
+        $payment_method_id = $payment_method_info[1];
```
Justintime50, line 63: I believe we can do this here:

```suggestion
        [$payment_method_endpoint, $payment_method_id] = Billing::get_payment_info($primary_or_secondary);
```

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+     * @param string $amount
+     * @param string $primary_or_secondary
+     * @param string $api_key
+     * @return mixed
+     */
+    public static function fund_wallet($amount, $primary_or_secondary = 'primary', $api_key = null)
+    {
+        $payment_method_info = Billing::get_payment_info($primary_or_secondary);
+        $payment_method_endpoint = $payment_method_info[0];
+        $payment_method_id = $payment_method_info[1];
```
Justintime50, line 43: I believe we can do this here:

```suggestion
        [$payment_method_endpoint, $payment_method_id] = Billing::get_payment_info($primary_or_secondary);
```

Justintime50 (changes requested):

lib/EasyPost/Billing.php
jchen293: resolved

lib/EasyPost/Billing.php
jchen293: resolved

lib/EasyPost/Billing.php
jchen293: resolved

lib/EasyPost/Billing.php
jchen293: resolved

lib/EasyPost/Billing.php
jchen293: resolved

lib/EasyPost/Billing.php
jchen293: resolved

nwithan8 (approved):

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+        ];
+        $payment_method_to_use = $payment_method_map[$primary_or_secondary] ?? null;
+
+        if ($payment_methods->$payment_method_to_use != null) {
+            $payment_method_id = $payment_methods->$payment_method_to_use->id;
+        }
+        echo "payment_method_id: " . $payment_method_id;
+        if ($payment_method_to_use !== null && $payment_method_id !== null) {
+            if (strpos($payment_method_id, 'card_') === 0) {
+                return ['/credit_cards', $payment_methods->$payment_method_to_use->id];
```
Justintime50, line 91: These return statements should instead be returning variables. Offhand, the reader doesn't know what `/credit_cards` is supposed to be as the first return value; however, if instead we returned `$endpoint`, it's easy to know. Additionally, we should only have one return value in here. Assign the values of the variables inside of the if/else statements and then return once. The throw error below should be in an `else` block.

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+            } else if (strpos($payment_method_id, 'bank_') === 0) {
+                echo "\npayment_method_id went in here ";
```
Justintime50, line 93: Never echo code in production. This needs to be removed.

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,100 @@
...
+                return ['/bank_accounts', $payment_methods->$payment_method_to_use->id];
+            }
+        }
+
+        throw new Error('The chosen payment method is not valid. Please try again.');
```
Justintime50, line 98: This should be an an `else` statement.

Justintime50 (changes requested):

lib/EasyPost/Billing.php
jchen293: removed.

lib/EasyPost/Billing.php
jchen293: resolved

lib/EasyPost/Billing.php
jchen293: resolved.

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,104 @@
...
+        $payment_methods = Billing::retrieve_payment_methods();
+        $payment_method_map = [
+            'primary' => 'primary_payment_method',
+            'secondary' => 'secondary_payment_method'
+        ];
+        $payment_method_to_use = $payment_method_map[$primary_or_secondary] ?? null;
+
+        if ($payment_methods->$payment_method_to_use != null) {
+            $payment_method_id = $payment_methods->$payment_method_to_use->id;
+        }
```
Justintime50, line 87: This needs an else statement where `$payment_method_id` has a fallback value, otherwise it will fail below when it tries to get the `strpos` of nothing.

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,104 @@
...
+
+        $endpoint = $payment_id = '';
+        if ($payment_method_to_use !== null && $payment_method_id !== null) {
+            if (strpos($payment_method_id, 'card_') === 0) {
+                $endpoint = '/credit_cards';
+                $payment_id = $payment_methods->$payment_method_to_use->id;
```
Justintime50, line 93: This line can be removed since we already defined it above as `$payment_method_id`.

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,104 @@
...
+            } else if (strpos($payment_method_id, 'bank_') === 0) {
+                $endpoint = '/bank_accounts';
+                $payment_id = $payment_methods->$payment_method_to_use->id;
```
Justintime50, line 96: This line can be removed since we already defined it above as `$payment_method_id`.

Justintime50 (changes requested):

lib/EasyPost/Billing.php
jchen293: fixed

lib/EasyPost/Billing.php
jchen293: fixed

lib/EasyPost/Billing.php
jchen293: good catch

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,103 @@
...
+        $payment_method_map = [
+            'primary' => 'primary_payment_method',
+            'secondary' => 'secondary_payment_method'
+        ];
+        $payment_method_to_use = $payment_method_map[$primary_or_secondary] ?? null;
+
+        if ($payment_methods->$payment_method_to_use->id != null) {
+            $payment_method_id = $payment_methods->$payment_method_to_use->id;
+        } else {
+            throw new Error('Did not find any chosen payment method. Please try again.');
```
Justintime50:
```suggestion
            throw new Error('Not a valid payment method. Please try again.');
```

lib/EasyPost/Billing.php
```diff
@@ -0,0 +1,103 @@
...
+    private static function get_payment_info($primary_or_secondary = 'primary')
+    {
+        $payment_methods = Billing::retrieve_payment_methods();
+        $payment_method_map = [
+            'primary' => 'primary_payment_method',
+            'secondary' => 'secondary_payment_method'
+        ];
+        $payment_method_to_use = $payment_method_map[$primary_or_secondary] ?? null;
+
+        if ($payment_methods->$payment_method_to_use->id != null) {
```
Justintime50:
```suggestion
        if ($payment_method_to_use != null && $payment_methods->$payment_method_to_use->id != null) {
```

Justintime50 (approved):

nwithan8 (approved):

Issue #181, opened 2022-06-15
[Bug]: EasyPost Errors are sometimes array type and not string


LoganSimonsen:
### Software Version

PhpClient/3.6.0

### Language Version

na

### Operating System

na

### What happened?

When parsing "errors" the library assumes that these errors will be strings, so a breaking error is thrown when "errors" is an array of strings. 

![image](https://user-images.githubusercontent.com/32521559/173905633-8710bb45-0df4-4cb3-a630-1270873c0710.png)

### What was expected?

handle EasyPost errors as a string or parse out the array of strings.

belisarh:
Hey all - any progress on this?

Justintime50:
@belisarh we plan to address this in the coming months. All of our libraries are getting a fresh coat of paint where error handling is concerned. We want to ensure consistency across libs so we are taking a holistic view at what will be needed.

PR #189, opened 2022-07-01
Adds "errors" key handling on error responses

belisarh:
# Description

- Fixes [https://github.com/EasyPost/easypost-php/issues/181](https://github.com/EasyPost/easypost-php/issues/181 ) by adding handling for cases where the error response message is found under the `errors` key. Currently only the `error` key is handled, leading to unhandled errors. 

# Testing
Not exactly sure if / how we can update the tests here. It seems only a specific error path is tested currently. This preserves current behavior by adding an `errors` key handling to handle GLS error responses from breaking apps when using the library. The other issue is that apparently production keys are required when running tests.

# Pull Request Type

### DIFF
--- a/lib/EasyPost/Requestor.php
+++ b/lib/EasyPost/Requestor.php
@@ -297,7 +297,7 @@ class Requestor
 
         $message = "";
 
-        if (is_array($response['error']) ) {
+        if (is_array($response['error'])) {
             $message = $response['error']['message'];
         } elseif (!empty($response['error'])) {
             $message = $response['error'];
--- a/lib/EasyPost/Requestor.php
+++ b/lib/EasyPost/Requestor.php
@@ -291,10 +291,25 @@ class Requestor
      */
     public function handleApiError($httpBody, $httpStatus, $response)
     {
-        if (!is_array($response) || !isset($response['error'])) {
+        if (!is_array($response) && (!isset($response['error']) || ! isset($response['errors']))) {
             throw new Error("Invalid response object from API: HTTP Status: ({$httpStatus}) {$httpBody})", $httpStatus, $httpBody);
         }
-        throw new Error(is_array($response['error']) ? $response['error']['message'] : (!empty($response['error']) ? $response['error'] : ""), $httpStatus, $httpBody);
+
+        $message = "";
+
+        if (is_array($response['error']) ) {
+            $message = $response['error']['message'];
+        } elseif (!empty($response['error'])) {
+            $message = $response['error'];
+        }
+
+        if (isset($response['errors']) && is_array($response['errors'])) {
+            $message = join(' ', $response['errors']);
+        } elseif (!empty($response['errors'])) {
+            $message = $response['errors'];
+        }
+
+        throw new Error($message, $httpStatus, $httpBody);
     }
 
     /**
--- a/lib/EasyPost/Requestor.php
+++ b/lib/EasyPost/Requestor.php
@@ -291,7 +291,7 @@ class Requestor
      */
     public function handleApiError($httpBody, $httpStatus, $response)
     {
-        if (!is_array($response) && (!isset($response['error']) || ! isset($response['errors']))) {
+        if (!is_array($response) || !isset($response['error'])) {
             throw new Error("Invalid response object from API: HTTP Status: ({$httpStatus}) {$httpBody})", $httpStatus, $httpBody);
         }
 


Justintime50 (changes requested):
As mentioned on #181 this is something we plan to correct holistically across all our libs and will contain more changes than those listed in this PR (note there are some additional similar code blocks in the `Error` file). If you need these changes immediately, I'd suggest building from source until we circle back to this (we have a couple other inflight projects at the moment). I'll be sure to update you here with our progress on overhauling error handling.

belisarh:
We will follow up on this via email rather than here.

lib/EasyPost/Requestor.php
```diff
@@ -294,7 +294,22 @@ public function handleApiError($httpBody, $httpStatus, $response)
...
+            $message = $response['error']['message'];
+        } elseif (!empty($response['error'])) {
+            $message = $response['error'];
+        }
+
+        if (isset($response['errors']) && is_array($response['errors'])) {
+            $message = join(' ', $response['errors']);
+        } elseif (!empty($response['errors'])) {
+            $message = $response['errors'];
+        }
```
Justintime50, line 310: Although this is the right idea, this doesn't work for a couple different reasons:

1. The errors are found at `$response['error']['errors']`
2. This will clobber the `$response['error']['message']` field since we set it just a couple lines before and are reassigning here. We should really be appending to that, if it happens to be empty there should be no harm to append since we set the message to an empty string at the start.
3. The `error['errors']` field is a list of objects. Simply joining here doesn't work because you can't join objects as a string. We'd need to serialize the errors here prior to joining them.

After double checking this with our plans, if we can correct the above I'll happily merge this. We have some additional tweaks to make in our next major release of this library; however, this bug fix shouldn't interfere with those planned changes and could go out now instead.

Justintime50 (changes requested):

lib/EasyPost/Requestor.php
belisarh, line 310: @Justintime50 Looking again at this and trying to figure out why this is failing on the current implementation.

It appears that the only possibility is that **you are getting an array under `$response['error']['message']`**, at least in some, if not all, GLS error responses. Other possibilities would be handled correctly by the current implementation.

In that case, the below change to the current implementation would fix the issue and preserve current behavior in all cases that don't trigger the bug.

```suggestion
        if (is_array($response['error'])) {
            $message = is_array($response['error']['message'])
                ? json_encode($response['error']['message'])
                : $response['error']['message'];
        } elseif (!empty($response['error'])) {
            $message = $response['error'];
        }
```

Justintime50: Interesting, can you share a complete error you've received from GLS? (the whole object), I believe you and I are talking about different kinds of errors (I'm referring to the more typical errors the API should be returning while I think you are referring to a GLS specific error that probably needs to be corrected at the API level instead of here as it may be putting that array of errors in the wrong field (which is then triggering this error). Seeing a complete example of a GLS error (and not the shipment create error I mentioned) would be helpful to determine that and next steps.

Justintime50: Thanks for confirming this. This is actually a bug at the API level (most of our carrier integrations pass string errors in the `message` field while arrays of errors get passed in the `errors` field. If this carrier had been following this convention, this error would not be thrown as the PHP library here was built with the assumption that the API would return the right data. I'm speaking with our internal teams about this to see if we can get this resolved for GLS (GSO). As such, I'm not too keen on integrating this one-off fix for this carrier's improper error handling if we can instead fix this at the API level instead. I'll keep you posted on what I hear from them. As mentioned before, if a speedy fix is necessary for your integration, I'd suggest patching this locally until we can correct this at the API level or determine we must fix it here.

Justintime50:
`v5.3.0` has been released with this patch, please upgrade and let me know if this corrects these issues with GSO/GLS. Thank you for your contribution here (it was used in another PR, I wasn't able to push to yours since you didn't branch and instead made changes on your master branch).

I apologize for the back and forth here and appreciate your patience. Initially the problem was misunderstood. EasyPost errors typically have a `message` field which is a string and then an `errors` field which is an array of errors. The initial ticket and naming of this PR made it sound like the problem resided in the `errors` field and not the `message` field which led to confusion and my line of questioning when instead one carrier was improperly passing an array back via the string field. We've put up this bandaid temporarily in this library and plan to correct this issue at the API level as well shortly. 

We still plan to make additional improvements for error handling in our client libraries in the coming couple months. This is a good first step.

PR #191, opened 2022-07-07
fix: coerce array errors to strings if present

Justintime50:
# Description

The changes from #189 are found here with additional comments and tests. Cannot push to the original PR as it was not created off a branch and was on the master branch. These changes are backwards compatible while providing a bandaid fix for carriers like GSO/GLS.

# Testing

- [x] Adds various tests to ensure we can pull out error properties properly
- [x] A/B tested an error message that was an array (placed the string into an array in the error cassette and it parsed properly)

# Pull Request Type

### DIFF
--- a/lib/EasyPost/Requestor.php
+++ b/lib/EasyPost/Requestor.php
@@ -294,7 +294,19 @@ class Requestor
         if (!is_array($response) || !isset($response['error'])) {
             throw new Error("Invalid response object from API: HTTP Status: ({$httpStatus}) {$httpBody})", $httpStatus, $httpBody);
         }
-        throw new Error(is_array($response['error']) ? $response['error']['message'] : (!empty($response['error']) ? $response['error'] : ""), $httpStatus, $httpBody);
+
+        // Errors from some carriers may be an array improperly assigned to the `message` field, check for those here (eg: GSO)
+        // TODO: This is a bandaid fix until the GSO service maps its error messages correctly. More dtails can be found here:
+        // https://github.com/EasyPost/easypost-php/pull/189
+        if (is_array($response['error'])) {
+            $message = is_array($response['error']['message'])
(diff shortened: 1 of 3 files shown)


jchen293 (approved):

lib/EasyPost/Requestor.php
```diff
@@ -294,7 +294,19 @@ public function handleApiError($httpBody, $httpStatus, $response)
...
             throw new Error("Invalid response object from API: HTTP Status: ({$httpStatus}) {$httpBody})", $httpStatus, $httpBody);
         }
-        throw new Error(is_array($response['error']) ? $response['error']['message'] : (!empty($response['error']) ? $response['error'] : ""), $httpStatus, $httpBody);
+
+        // Errors from some carriers may be an array improperly assigned to the `message` field, check for those here (eg: GSO)
+        // TODO: This is a bandaid fix until the GSO service maps its error messages correctly. More dtails can be found here:
+        // https://github.com/EasyPost/easypost-php/pull/189
+        if (is_array($response['error'])) {
+            $message = is_array($response['error']['message'])
+                ? json_encode($response['error']['message'])
```
nwithan8, line 303: We're sure the data will be JSON? What if this throws an error?

nwithan8 (approved):

lib/EasyPost/Requestor.php
Justintime50: This simply returns a string of the serialized data it gets: https://www.php.net/manual/en/function.json-encode.php. Handles strings, objects, arrays:

```
string(7) ""hello world""
string(17) "{"hello":"world"}"
string(17) "["hello","world"]"
```

Basically if the error is an array (should only be this single carrier), it will go into this block, otherwise it'll do exactly what it's always done (and what it should be doing).

PR #218, opened 2022-09-15
feat: swaps the underlying curl lib for guzzle

Justintime50:
# Description

- Adopts `Guzzle` as the HTTP client for this library. This should provide a much more consistent experience, better encoding, faster request times in some cases. By using Guzzle we can remove the old SSL logic and completely rip out the curl logic we homegrew.

# Testing

- [x] Fixed various tests with new encoding, new rates, etc
- [x] E2E tested timeouts, failures to connect without a connection, etc
- [x] Re-recorded all cassettes to use the new HTTP client and catch various edge cases

# Pull Request Type


nwithan8 (approved):

Justintime50:
Drafting till I can merge the other open PR then rebase this one

lib/EasyPost/EasyPost.php
```diff
@@ -118,50 +109,26 @@ public static function setApiVersion($apiVersion)
...
-    }
-
-    /**
-     * Set time in milliseconds to wait for a connection.
-     *
-     * Zero or null means no timeout.
-     *
-     * @param int|null $connectTimeout
-     */
-    public static function setConnectTimeout($connectTimeout)
```
jchen293, line 139: Why is `setConnectTimeout` removed? I assume Guzzle doesn't support this?

jchen293 (approved):

lib/EasyPost/EasyPost.php
Justintime50: Correct, they only have one timeout param instead of allowing for request and connection.

PR #232, opened 2022-11-23
Switch endpoint for carrier account creation depending on type

nwithan8:
# Description

- Switch endpoint for carrier account creation depending on type

# Testing

- New unit test, cassette

# Pull Request Type


lib/EasyPost/CarrierAccount.php
```diff
@@ -76,7 +77,14 @@ public static function create($params = null, $apiKey = null)
             unset($params);
             $params['carrier_account'] = $clone;
         }
-        return self::createResource(get_class(), $params, $apiKey);
+        $type = $params['carrier_account']['type'];
+        if (is_null($type)) {
```
Justintime50, line 81: For PHP, you'll want this instead. Otherwise we won't get the intended affect:

```suggestion
        if (!isset($type)) {
```

lib/EasyPost/CarrierAccount.php
```diff
@@ -76,7 +77,14 @@ public static function create($params = null, $apiKey = null)
...
+            throw new Error('type is required');
+        }
+        $url = self::selectCarrierAccountCreationEndpoint($type);
+        $requestor = new Requestor($apiKey);
+        list($response, $apiKey) = $requestor->request('post', $url, $params);
+        return Util::convertToEasyPostObject($response, $apiKey);
```
Justintime50, line 87: Let's use logical line breaks to split up this wall of text here.

lib/EasyPost/CarrierAccount.php
```diff
@@ -92,4 +100,13 @@ public static function types($params = null, $apiKey = null)
         list($response, $apiKey) = $requestor->request('get', '/carrier_types', $params);
         return Util::convertToEasyPostObject($response, $apiKey);
     }
+
+    private static function selectCarrierAccountCreationEndpoint($carrierAccountType): string
```
Justintime50, line 104: Please add a docblock like all other functions.

lib/EasyPost/Constants.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+namespace EasyPost;
+
+abstract class Constants
+{
+    const CARRIER_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS = [
```
Justintime50, line 7: We shouldn't need to be defining abstract classes for constants. Constants should sit at the root level and be imported where needed. Wrapping them in a class only abstracts them away further without any benefit.

In the case of PHP, they need to be at the root level:

Screen Shot 2022-11-23 at 3 38 41 PM
https://www.php.net/manual/en/language.constants.syntax.php

Justintime50 (changes requested):

Justintime50:
This needs a rebase

test/EasyPost/CarrierAccountTest.php
```diff
@@ -3,6 +3,10 @@
 namespace EasyPost\Test;
 
 use EasyPost\CarrierAccount;
+use EasyPost\EasyPost;
+use EasyPost\FieldError;
+use EasyPost\Test\Fixture;
+use VCR\VCR;
```
Justintime50, line 9: This isn't using the new `setupCassette` below. When this is done, it won't require any of these.

```suggestion
use EasyPost\Test\Fixture;
```

test/EasyPost/CarrierAccountTest.php
```diff
@@ -38,6 +42,42 @@ public function testCreate()
         $carrierAccount->delete(); // Delete the carrier account once it's done being tested.
     }
 
+    /**
+     * Test creating a carrier account with custom workflow.
+     */
+    public function testCreateWithCustomWorkflow()
+    {
+        VCR::insertCassette('carrier_accounts/create_with_custom_workflow.yml');
```
Justintime50, line 50: Please use the new `TestUtil::setupCassette()` function instead:

```suggestion
        TestUtil::setupCassette('carrier_accounts/createWithCustomWorkflow.yml');
```

The filename also needs to follow convention

Justintime50 (changes requested):

test/EasyPost/CarrierAccountTest.php
```diff
@@ -3,6 +3,7 @@
 namespace EasyPost\Test;
 
 use EasyPost\CarrierAccount;
+use EasyPost\Test\Fixture;
```
Justintime50, line 6: This import isn't needed because it's already in the namespace (eg: `Test`)

test/EasyPost/CarrierAccountTest.php
```diff
@@ -38,6 +39,42 @@ public function testCreate()
...
+        TestUtil::setupCassette('carrier_accounts/createWithCustomWorkflow.yml');
+
+        $data = Fixture::basicCarrierAccount();
+        $data['type'] = 'FedexAccount';
+        // We have to send some registration data, otherwise API will throw a 400 Bad Request.
+        $data['registration_data'] = [
+            'some' => 'data',
+        ];
+
+        // catch exception
```
Justintime50, line 56: There's no need to comment `catch exception` since we... catch the exception.

test/EasyPost/CarrierAccountTest.php
```diff
@@ -38,6 +39,42 @@ public function testCreate()
...
+        try {
+            $carrierAccount = CarrierAccount::create($data);
+            // Delete the carrier account once it's done being tested (should not be reached).
+            $carrierAccount->delete();
+        } catch (\EasyPost\Error $e) {
+            $this->assertEquals(422, $e->getHttpStatus());
+            $this->assertNotEmpty($e->errors);
```
Justintime50, line 63: I have reminded various times now that checking that something isn't empty right before a for loop to check individual records isn't necessary because the following assertions in the for loop would fail if something was necessary.

Please remember this feedback moving forward so I don't need to keep reminding and remove this assertion here
```suggestion
```

test/EasyPost/CarrierAccountTest.php
```diff
@@ -38,6 +39,42 @@ public function testCreate()
...
+            $errorFound = false;
+            $errors = $e->errors;
+            foreach ($errors as $error) {
+                if ($error['field'] == 'account_number' && $error['message'] == 'must be present and a string') {
+                    $errorFound = true;
+                    unset($error);
```
Justintime50, line 69: Why are we unsetting the error here? There should be no need since we immediately break and assert something else

test/EasyPost/CarrierAccountTest.php
```diff
@@ -38,6 +39,42 @@ public function testCreate()
...
+                    break;
+                }
+                unset($error);
```
Justintime50, line 72: Why are we unsetting the error here, again? There should be no need since we immediately break and assert something else

test/EasyPost/CarrierAccountTest.php
```diff
@@ -38,6 +39,42 @@ public function testCreate()
...
     }
 
+    /**
+     * Test creating a carrier account with custom workflow.
+     */
+    public function testCreateWithCustomWorkflow()
+    {
+        TestUtil::setupCassette('carrier_accounts/createWithCustomWorkflow.yml');
+
+        $data = Fixture::basicCarrierAccount();
```
Justintime50, line 49: One item I see wrong (check the cassette) is that you are reusing this fixture which we don't want in this test. It's including all the extra `credentials` and `test_credentials` fields which may be confusing this endpoint. We simply want to create a new object with `type` and `registration_data` fields only.

Justintime50 (changes requested):
CI is failing because you didn't rename the cassette when you changed the name of it in the test so it couldn't find it, tried to make a real request, and then it failed without creds.

With that said, please remove the cassette completely and re-record because the fixture data used was incorrect.

Justintime50 (approved):

PR #233, opened 2022-11-23
fix: api key retrieval logic, tests, and scrubbing

Justintime50:
# Description

For the longest time, we've been uncertain as to what this for loop did in the API key retrieval function. Figured it out - it let's a parent retrieve the API keys of a child user without needing to authenticate as the child user by filtering out all other records.

This PR adds better comments and a test for this functionality.

# Testing
Adds a new unit test and re-records older API key tests

# Pull Request Type


lib/EasyPost/User.php
```diff
@@ -97,36 +97,36 @@ public static function all_api_keys($apiKey = null)
...
 
-        if (is_null($myApiKeys)) {
-            return null;
-        } else {
-            $response = [];
-            foreach ($myApiKeys as $key) {
-                $response[$key->mode] = $key->key;
-            }
-            return $response;
+        // TODO: Don't rewrap objects here, return the direct response objects
```
jchen293, line 123: Question: This TODO and the TODOs in the unit tests are for the PHP thread-safe rewrite?

jchen293 (approved):

PR #237, opened 2022-12-02
fix!: camelcase for all function names and params

Justintime50:
# Description

Uses the proper PHP syntax by switching function names and params to camel case making everything in this lib uniform and enforced without exceptions.

# Testing

Tests will catch anything not setup correctly

# Pull Request Type


nwithan8:
Can you provide a source for this style rule? SO posts suggest there is no standard, and [this PHP guide](https://www.php.net/manual/en/userlandnaming.rules.php) actually suggest functions should be snake case rather than camel case.

Justintime50:
> Can you provide a source for this style rule? SO posts suggest there is no standard, and [this PHP guide](https://www.php.net/manual/en/userlandnaming.rules.php) actually suggest functions should be snake case rather than camel case.

There is no true standard in PHP - it's one or the others. "PHP the language" uses underscores; however, most of the large frameworks and libraries in the PHP ecosystem instead use camelCase. We had a mix and needed to unify it. 

The ruleset that we use (what the linter uses) in our PHP lib is [PSR12](https://www.php-fig.org/psr/psr-12/) which does state function and param names must be camelCase (no direct link but lots of examples) and [Laravel](https://webdevetc.com/blog/laravel-naming-conventions/) for instance (biggest framework in the space) uses camelCase as well.

nwithan8 (approved):

jchen293 (approved):

PR #239, opened 2022-12-06
feat: thread safety via new EasyPostClient object

Justintime50:
# Description

Uses new EasyPostClient object which provides thread-safety. All functions are now called against a client object ensuring that API keys cannot get mixed via the global state. Syntax now looks like `$client->address->create();`. Various files were moved to provide better project structure, a few functions were shuffled, lots of code was removed as it was no longer necessary with the architecture changes here. Code that has not been confirmed dead yet was retained for now (can revisit and remove later, most of it lives in `EasyPostObject`).

# Testing

- All the tests were updated to reflect the new syntax
- Tested autocomplete does indeed work (won't work inside our test suite but did with a scratch file)
- The `require` test did a great job of reminding me to catch all the new imports needed

# Pull Request Type


lib/EasyPost/Services/CarrierAccountService.php
```diff
@@ -0,0 +1,115 @@
...
+use EasyPost\EasyPostObject;
+use EasyPost\Error;
+use EasyPost\Requestor;
+
+/**
+ * CarrierAccount service containing all the logic to make API calls.
+ */
+class CarrierAccountService extends BaseService
+{
+    protected static $modelClass = 'CarrierAccount';
```
jchen293, line 15: This can be private

test/EasyPost/AddressTest.php
```diff
@@ -30,7 +30,9 @@ public function testCreate()
     {
         TestUtil::setupCassette('addresses/create.yml');
 
-        $address = Address::create(Fixture::caAddress1());
+        $client = new EasyPostClient(getenv('EASYPOST_TEST_API_KEY'));
```
jchen293, line 33: This could be a global variable so we don't have to define it every time for each unit test

lib/EasyPost/Services/BaseService.php
```diff
@@ -0,0 +1,189 @@
...
+use EasyPost\Util;
+
+class BaseService
+{
+    /**
+     * Service constructor shared by all child services.
+     *
+     * @param EasyPostClient $client
+     */
+    public function __construct($client)
```
jchen293, line 17: Since the constructor here is a `public`, end users will be able to initialize each service?

test/EasyPost/AddressTest.php
Justintime50: Yeah, I'll make sure to revisit this with some kind of utility once done.

lib/EasyPost/Services/CarrierAccountService.php
Justintime50: Whoops, nice catch!

lib/EasyPost/Services/BaseService.php
Justintime50: Yeah, I'm getting errors if it's not set to public... how can we fix this?

Justintime50: I believe this worked, I thought it'd get mad. We'll see if this continues.

test/EasyPost/AddressTest.php
Justintime50: Fixed

lib/EasyPost/EasyPostObject.php
```diff
@@ -42,33 +30,17 @@ class EasyPostObject implements \ArrayAccess, \Iterator
...
-     * @param string $apiKey
+     * @param EasyPostClient $client
      * @param string $parent
      * @param string $name
      */
-    public function __construct($id = null, $apiKey = null, $parent = null, $name = null)
+    public function __construct($client, $parent = null, $name = null)
     {
-        $this->_apiKey = $apiKey;
+        $this->_client = $client;
```
nwithan8, line 39: So we are going to pass the client down into objects?

Justintime50: Actually this is leftover from when I first started and isn't even used. I've removed this now.

The `BaseService` does accept the client object so that we can use the API key in requests, each Service then inherits the BaseService. I believe this is the same pattern we used in Java correct? But no, actual "EasyPost" objects should not have the client associated with the client itself.

nwithan8: Correct, a Service accepts a client, a Object does not. Please remove.

lib/EasyPost/Service/BaseService.php
```diff
@@ -0,0 +1,182 @@
...
+
+    /**
+     * Internal delete method.
+     *
+     * @param string $class
+     * @param string $id
+     * @param mixed $params
+     * @return void
+     */
+    protected function deleteResource($class, $id, $params = null)
```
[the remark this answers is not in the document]
Justintime50: Nice catch, fixed.

lib/EasyPost/Http/Requestor.php
```diff
@@ -32,12 +25,12 @@ public function __construct($apiKey = null)
      * @param string $url
      * @return string
      */
-    public static function apiUrl($url = '', $beta = false)
+    public function absoluteUrl($url = '', $beta = false)
     {
         if ($beta) {
-            $apiBase = EasyPost::$betaApiBase;
+            $apiBase = Constant::BETA_API_BASE;
```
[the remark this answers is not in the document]
Justintime50: 🤷‍♂️ this is how it's been. I can swap it out for that though.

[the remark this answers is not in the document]
Justintime50: 🤷‍♂️ It's been this way, just copied it over. I tweaked what you asked for though as that's probably more correct. "Base" shouldn't include the version or beta

Justintime50:
> Can we group all the models (`Shipment.php`, `Rate.php`, etc.) into a "Model" folder?

Unfortunately not without undesirable consequences. In PHP, the folder is important to the namespace and we don't want root level "EasyPost" objects to be nestled in some "Model" folder, asserting they are a type of Address would no longer be `/EasyPost/Address` it would be an awkward `EasyPost\Model\Address`. If you look at Stripe's PHP lib, they've done the same actually. I've pulled everything that isn't a "Model" into a separate namespace/folder to try and keep them consistent that way.

nwithan8 (approved):

lib/EasyPost/EasyPostClient.php
```diff
@@ -0,0 +1,100 @@
...
+        $this->report = new ReportService($this);
+        $this->scanForm = new ScanFormService($this);
+        $this->shipment = new ShipmentService($this);
+        $this->tracker = new TrackerService($this);
+        $this->user = new UserService($this);
+        $this->webhook = new WebhookService($this);
+
+        if (!$this->apiKey) {
+            throw new Error('No API key provided. See https://www.easypost.com/docs for details, or contact ' . Constant::SUPPORT_EMAIL . ' for assistance.');
+        }
```
jchen293, line 98: Nit: this could be on the top before the initialization of the services

lib/EasyPost/Http/Requestor.php
```diff
@@ -32,12 +25,12 @@ public function __construct($apiKey = null)
      * @param string $url
```
jchen293, line 25: missing docstring for `$beta` in the below functions

lib/EasyPost/Constant/Constant.php
```diff
@@ -0,0 +1,22 @@
+<?php
+
+namespace EasyPost\Constant;
+
+abstract class Constant
```
jchen293, line 5: This class name should be plural `constants` to match Java and C#

lib/EasyPost/Service/AddressService.php
```diff
@@ -0,0 +1,99 @@
...
+     */
+    public function createAndVerify($params = null)
+    {
+        if (!isset($params['address']) || !is_array($params['address'])) {
+            $clone = $params;
+            unset($params);
+            $params['address'] = $clone;
+        }
+
+        $requestor = new Requestor($this->client);
```
jchen293, line 78: Question: have we thought about adding another parameter for `client` in the `request` function and making it a static function, so we could avoid this object initialization every time?

lib/EasyPost/Util/Util.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Validate a webhook originated from EasyPost by comparing the HMAC header to a shared secret.
+     * If the signatures do not match, an error will be raised signifying the webhook either did not originate
+     * from EasyPost or the secrets do not match. If the signatures do match, the `event_body` will be returned as JSON.
+     *
+     * @param mixed $eventBody
+     * @param mixed $headers
+     * @param string $webhookSecret
+     * @return mixed
```
jchen293, line 84: Missing docstring for `throws error`

lib/EasyPost/Util/Util.php
```diff
@@ -0,0 +1,128 @@
...
+
+    /**
+     * Get the lowest SmartRate from a list of SmartRates.
+     *
+     * To exclude a carrier or service, prepend the string with `!`.
+     *
+     * @param array $smartRates
+     * @param int $deliveryDays
+     * @param string $deliveryAccuracy
+     * @return Rate
```
jchen293, line 41: Missing docstring for `throws error`

CHANGELOG.md
```diff
@@ -3,6 +3,17 @@
...
+  - All instance methods and the ability to have objects update in-place has been removed due to this rearchitecture. When updating objects or performing actions on them (eg: creating a label/scanform), you will want to assign the return value to a variable and use that moving forward
+  - Functions no longer accept an API key as an optional param
+- `EasyPostResource` is now `BaseService`
+- Models and Services
+- `Requestor` has moved to `Http`, variables from `EasyPost` now moved to `Constant`, `Error` moved to `Exception`
+- `->save()` instance methods are now `update()` static methods
+- Occurances of `smartrate` are now `smartRate` and `Smartrate` are now `SmartRate` to match the documentation and API expectations
+- `Referral` is now `ReferralCustomer` to better match documentation and API expectation
+- `validateWebhook`, `getLowestSmartRate`, and `receiveEvent` are now under `EasyPost\Util\Util` as they do not make any API calls
+  - Internal only utilities have been moved to `EasyPost\Util\InternalUtil`
```
jchen293, line 16: We should call out that the `endshipper` in beta is removed

jchen293 (commented):
Code looks good, just a few questions

CHANGELOG.md
Justintime50: Sure

lib/EasyPost/EasyPostClient.php
Justintime50: You need to initialize an object before you can perform logic on said object. This is a pretty traditional flow so I feel how it currently is makes the most sense.

jchen293: Either way works, but I feel there is no point of creating services if the input API key is empty or null

lib/EasyPost/Service/AddressService.php
Justintime50: Normally that would be a nice way of cleaning things up; however, we use the client object throughout the Requestor class for various other things (timeout, apiKey, etc) and so that would only introduce the same amount of work in a different place. I think keeping it how it is makes sense for now.

lib/EasyPost/EasyPostClient.php
Justintime50: Sure, though it takes all of ~2ms to do so maybe so no harm no foul by waiting.

lib/EasyPost/Constant/Constant.php
nwithan8: > Class names must always be singular. 
While I mostly agree, we're not really using Constant(s) here as a class, but rather as a category to store values in, so I'm not sure that styling rule applies.

nwithan8 (approved):

lib/EasyPost/Constant/Constant.php
Justintime50: Fair enough. Since you both want plural for this and to Nate's point it's not an actual class, I'll update it.

lib/EasyPost/Service/AddressService.php
jchen293: If a user is making thousands of API calls in a day, creating `Requestor` object every time is not ideal for memory allocation. One of the reasons we initialize every service with `client`, when EasypPostClient is created, is to avoid another object initialization. Won't block on this, but we should come back for optimization in the future.

lib/EasyPost/Http/Requestor.php
```diff
@@ -1,43 +1,37 @@
...
 
     /**
      * Get the API URL.
      *
      * @param string $url
+     * @param bool $beta
      * @return string
      */
-    public static function apiUrl($url = '', $beta = false)
+    public function absoluteUrl($url = '', $beta = false)
```
jchen293, line 29:
```suggestion
    private function absoluteUrl($url = '', $beta = false)
```
This function should be `private`, it's only being used in `Requestor` file

jchen293 (approved):

lib/EasyPost/Http/Requestor.php
Justintime50: Nice catch, fixed

jchen293 (approved):

nwithan8 (approved):

PR #246, opened 2022-12-15
New test utility to mock HTTP requests/responses

nwithan8:
# Description

Based on the implementation in [C#](https://github.com/EasyPost/easypost-csharp/blob/master/EasyPost.Tests/ServicesTests/BillingServiceTest.cs#L21), [Go](https://github.com/EasyPost/easypost-go/pull/137) and [Node](https://github.com/EasyPost/easypost-node/pull/317), this adds the ability to mock requests.

This is accomplished by adding a `MockUtility` object into the EasyPost client. If this property is set, the client will ONLY use mock requests (any request passed in that is not a configured mock request will throw an exception).

# Testing
- Use mocking in new billing tests

Tests:
image
Coverage:
image

# Pull Request Type

### DIFF
--- a/README.md
+++ b/README.md
@@ -128,20 +128,15 @@ Some tests may require an EasyPost user with a particular set of enabled feature
 
 #### Mocking
 
-Some of our unit tests require HTTP calls that cannot be easily tested with live/recorded calls (e.g. HTTP calls that
-trigger payments or interact with external APIs).
+Some of our unit tests require HTTP calls that cannot be easily tested with live/recorded calls (e.g. HTTP calls that trigger payments or interact with external APIs).
 
-We have implemented a custom, lightweight HTTP mocking functionality in this library that allows us to mock HTTP calls
-and responses.
+We have implemented a custom, lightweight HTTP mocking functionality in this library that allows us to mock HTTP calls and responses.
 
 A mock client is the same as a normal client, with a set of mock request-response pairs stored as a property.
 
-At the time to make a real HTTP request, a mock client will instead check which mock request entry matches the queued
-request (matching by HTTP method and a regex pattern for the URL), and will return the corresponding mock response (HTTP
-status code and body).
+At the time of making a real HTTP request, a mock client will instead check which mock request entry matches the queued request (matching by HTTP method and a regex pattern for the URL), and will return the corresponding mock response (HTTP status code and body).
 
-**NOTE**: If a client is configured with a mocking utility, it will ONLY make mock requests. If it attempts to
-make a request that does not match any of the configured mock requests, the request will fail and trigger an exception.
+**NOTE**: If a client is configured with a mocking utility, it will ONLY make mock requests. If it attempts to make a request that does not match any of the configured mock requests, the request will fail and trigger an exception.
 
 To use the mocking utility:
 
--- a/README.md
+++ b/README.md
@@ -125,3 +125,50 @@ Some tests may require an EasyPost user with a particular set of enabled feature
 - `USPS_CARRIER_ACCOUNT_ID` (eg: one-call buying a shipment for non-EasyPost employees)
 - `PARTNER_USER_PROD_API_KEY` (eg: creating a referral user)
 - `REFERRAL_USER_PROD_API_KEY` (eg: adding a credit card to a referral user)
+
+#### Mocking
+
+Some of our unit tests require HTTP calls that cannot be easily tested with live/recorded calls (e.g. HTTP calls that
+trigger payments or interact with external APIs).
+
+We have implemented a custom, lightweight HTTP mocking functionality in this library that allows us to mock HTTP calls
+and responses.
+
+A mock client is the same as a normal client, with a set of mock request-response pairs stored as a property.
+
+At the time to make a real HTTP request, a mock client will instead check which mock request entry matches the queued
+request (matching by HTTP method and a regex pattern for the URL), and will return the corresponding mock response (HTTP
+status code and body).
+
+**NOTE**: If a client is configured with a mocking utility, it will ONLY make mock requests. If it attempts to
+make a request that does not match any of the configured mock requests, the request will fail and trigger an exception.
+
+To use the mocking utility:
+
+```php
+use EasyPost\Test\mocking\MockingUtility;
+use EasyPost\Test\mocking\MockRequest;
+use EasyPost\Test\mocking\MockRequestMatchRule;
+use EasyPost\Test\mocking\MockRequestResponseInfo;
+
+// create a mocking utility with a list of mock request-response pairs
+$mockingUtility = new MockingUtility(
+    new MockRequest(
+        new MockRequestMatchRule(
+            // HTTP method and regex pattern for the URL must both pass for the request to match
+            'post',
+            '/v2\\/bank_accounts\\/\\S*\\/charges$/'
+        ),
+        new MockRequestResponseInfo(
+            // HTTP status code and body to return when this request is matched
+            200,
+            '{}'
+        )
+    )
+);
+
+// create a new client with the mock requests
+$client = new EasyPostClient("some_key", Constants::TIMEOUT, Constants::API_BASE, $mockUtility);
+
+// use the client as normal
+```
--- a/README.md
+++ b/README.md
@@ -164,10 +164,11 @@ $mockingUtility = new MockingUtility(
             200,
             '{}'
         )
-    )
+    ),
+    ... // more mock requests
 );
 
-// create a new client with the mock requests
+// create a new client with the mocking utility
 $client = new EasyPostClient("some_key", Constants::TIMEOUT, Constants::API_BASE, $mockUtility);
 
 // use the client as normal
--- a/lib/EasyPost/EasyPostClient.php
+++ b/lib/EasyPost/EasyPostClient.php
@@ -59,6 +59,7 @@ class EasyPostClient extends BaseService
     private $apiKey;
     private $timeout;
     private $apiBase;
+    private $mockingUtility;
 
     // Services
     public $address;
@@ -89,13 +90,15 @@ class EasyPostClient extends BaseService
      * @param string $apiKey
      * @param float $timeout
      * @param string $apiBase
+     * @param object $mockingUtility
      */
-    public function __construct($apiKey, $timeout = Constants::TIMEOUT, $apiBase = Constants::API_BASE)
+    public function __construct($apiKey, $timeout = Constants::TIMEOUT, $apiBase = Constants::API_BASE, $mockingUtility = null)
     {
         // Client properties
         $this->apiKey = $apiKey;
         $this->timeout = $timeout;
         $this->apiBase = $apiBase;
+        $this->mockingUtility = $mockingUtility;
 
         // Services
         // TODO: Make these all read only when we support PHP >= 8.1
@@ -154,5 +157,25 @@ class EasyPostClient extends BaseService
     public function getApiBase()
     {
         return $this->apiBase;
+    }
+
+    /**
+     * Get whether this client is configured to mock requests
+     *
+     * @return bool
+     */
+    public function mock()
+    {
+        return $this->mockingUtility !== null;
+    }
+
+    /**
+     * Get the mock requests of an EasyPostClient.
+     *
+     * @return object
+     */
+    public function getMockingUtility()
+    {
+        return $this->mockingUtility;
     }
 }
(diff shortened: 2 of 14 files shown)


lib/EasyPost/Http/Requestor.php
```diff
@@ -178,6 +178,20 @@ private static function requestRaw($client, $method, $url, $params, $beta = fals
             'User-Agent' => 'EasyPost/v2 PhpClient/' . Constants::LIBRARY_VERSION . " PHP/$phpVersion OS/$osType OSVersion/$osVersion OSArch/$osArch",
         ];
 
+        if ($client->mock()) {
+            // If there are mock requests set, this client will ONLY make mock requests
+            $mockingUtility = $client->getMockingUtility();
+            $matchingRequest = $mockingUtility->findMatchingMockRequest($method, $absoluteUrl);
+            if ($matchingRequest === null) {
+                throw new HttpException("No matching mock request found for {$method} {$absoluteUrl}");
```
Justintime50, line 186: This error message needs to be a constant

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+
+namespace EasyPost;
+
+use EasyPost\Constant\Constants;
+use EasyPost\Exception\Api\PaymentException;
+use EasyPost\Exception\General\MissingParameterException;
+use EasyPost\Test\mocking\MockingUtility;
+use EasyPost\Test\mocking\MockRequest;
+use EasyPost\Test\mocking\MockRequestMatchRule;
+use EasyPost\Test\mocking\MockRequestResponseInfo;
```
Justintime50, line 11: PHP folders must be title cased.

README.md
```diff
@@ -125,3 +125,50 @@ Some tests may require an EasyPost user with a particular set of enabled feature
 - `USPS_CARRIER_ACCOUNT_ID` (eg: one-call buying a shipment for non-EasyPost employees)
 - `PARTNER_USER_PROD_API_KEY` (eg: creating a referral user)
 - `REFERRAL_USER_PROD_API_KEY` (eg: adding a credit card to a referral user)
+
+#### Mocking
+
+Some of our unit tests require HTTP calls that cannot be easily tested with live/recorded calls (e.g. HTTP calls that
```
Justintime50, line 131: As mentioned before, please match the current styling of the document which is not to break up paragraphs on newlines if they are too long.

README.md
```diff
@@ -125,3 +125,51 @@ Some tests may require an EasyPost user with a particular set of enabled feature
...
+trigger payments or interact with external APIs).
+
+We have implemented a custom, lightweight HTTP mocking functionality in this library that allows us to mock HTTP calls
+and responses.
+
+A mock client is the same as a normal client, with a set of mock request-response pairs stored as a property.
+
+At the time to make a real HTTP request, a mock client will instead check which mock request entry matches the queued
```
Justintime50, line 139:
```suggestion
At the time of making a real HTTP request, a mock client will instead check which mock request entry matches the queued
```

README.md
```diff
@@ -125,3 +125,51 @@ Some tests may require an EasyPost user with a particular set of enabled feature
...
+**NOTE**: If a client is configured with a mocking utility, it will ONLY make mock requests. If it attempts to
+make a request that does not match any of the configured mock requests, the request will fail and trigger an exception.
+
+To use the mocking utility:
+
+```php
+use EasyPost\Test\mocking\MockingUtility;
+use EasyPost\Test\mocking\MockRequest;
+use EasyPost\Test\mocking\MockRequestMatchRule;
+use EasyPost\Test\mocking\MockRequestResponseInfo;
```
Justintime50, line 152: Must be `Mocking`

lib/EasyPost/EasyPostClient.php
```diff
@@ -89,13 +90,15 @@ class EasyPostClient extends BaseService
      * @param string $apiKey
      * @param float $timeout
      * @param string $apiBase
+     * @param object $mockingUtility
      */
-    public function __construct($apiKey, $timeout = Constants::TIMEOUT, $apiBase = Constants::API_BASE)
+    public function __construct($apiKey, $timeout = Constants::TIMEOUT, $apiBase = Constants::API_BASE, $mockingUtility = null)
```
Justintime50, line 95: Shouldn't the default be `false`?

lib/EasyPost/EasyPostClient.php
```diff
@@ -155,4 +158,24 @@ public function getApiBase()
...
     }
+
+    /**
+     * Get whether this client is configured to mock requests
+     *
+     * @return bool
+     */
+    public function mock()
+    {
+        return $this->mockingUtility !== null;
```
Justintime50, line 169: I would expect we'd just to make sure this isn't false.

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+        // api key does not matter for mocking
+        self::$client = self::getClient($mockingUtility);
+    }
+
+    /**
+     * Cleanup the testing environment once finished.
+     */
+    public static function tearDownAfterClass(): void
+    {
+    }
```
Justintime50, line 85: This code does nothing, please remove it.

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+
+    /**
+     * Test that the billing service can retrieve payment methods.
+     *
+     * @param MockingUtility|null $mockUtility
+     * @return EasyPostClient
+     * @throws MissingParameterException
+     */
+    public static function getClient(MockingUtility $mockUtility = null): \EasyPost\EasyPostClient
```
Justintime50, line 94: We don't use type hints in this library yet, please don't add them one-off. 

Additionally, we need to `use` or import everything at the top level. This helps readers know exactly what is used in this file right when opening it. We shouldn't mix imports at the top and explicit calls inline later down the file.

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+    {
+        return new EasyPostClient(getenv('EASYPOST_TEST_API_KEY'), Constants::TIMEOUT, Constants::API_BASE, $mockUtility);
+    }
```
Justintime50, line 97: The docstring for this was copy/pasted and never updated.

Even still, this function isn't needed if convention is followed. PHPUnit already allows for a `setUpBeforeClass` function which is where you should be setting up the global client for this test file. Please follow the pattern established for all the previous tests.

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+use EasyPost\Test\mocking\MockRequestMatchRule;
+use EasyPost\Test\mocking\MockRequestResponseInfo;
+
+class BillingTest extends \PHPUnit\Framework\TestCase
+{
+    private static $client;
+
+    /**
+     * Set up the testing environment for this file.
+     * @throws MissingParameterException
```
Justintime50, line 19: There is no need to annotate tests with `throws` tags

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+    public static function getClient(MockingUtility $mockUtility = null): \EasyPost\EasyPostClient
+    {
+        return new EasyPostClient(getenv('EASYPOST_TEST_API_KEY'), Constants::TIMEOUT, Constants::API_BASE, $mockUtility);
+    }
+
+    public function testFundWallet()
+    {
+        try {
+            self::$client->billing->fundWallet('2000', 'primary');
+            $this->assertTrue(true);
```
Justintime50, line 103: Question: I'd expect we could assert the URL getting hit here. How would this ever throw an error?

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+        } catch (\Exception $exception) {
+            $this->fail('Exception thrown when we expected no error');
+        }
+    }
+
+    public function testFundWalletWithNoPriorityLevel()
```
Justintime50, line 109: These functions all need docblocks

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+        }
+    }
+
+    /**
+     * @throws PaymentException
+     */
+    public function testRetrievePaymentMethodSummary()
+    {
+        $paymentMethodSummary = self::$client->billing->retrievePaymentMethods();
+        $this->assertNotNull($paymentMethodSummary);
```
Justintime50, line 125: Separate code from assertions with a line break.

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+        try {
+            self::$client->billing->fundWallet('2000');
+            $this->assertTrue(true);
+        } catch (\Exception $exception) {
+            $this->fail('Exception thrown when we expected no error');
+        }
+    }
+
+    /**
+     * @throws PaymentException
```
Justintime50, line 120: Replace throws with a human readable sentence describing what this test does, see other libs for an example.

This is true of all the docblocks here. We shouldn't annotate errors thrown in tests because they could change and it's misleading because it tells the user it's important to this test when it isn't. Additionally, this test function does not throw anything at all. Only functions that physically throw an exception should be annotated as such.

test/EasyPost/BillingTest.php
```diff
@@ -0,0 +1,254 @@
...
+            $this->assertTrue(true);
+        } catch (\Exception $exception) {
+            $this->fail('Exception thrown when we expected no error');
+        }
+    }
+
+    public function testGetPaymentMethodPrioritySwitchCase()
+    {
+        // Deleting a payment method gets the payment method internally, which should test the switch case.
+        // The payment method is not exposed by this method, so we can't assert against it. If this test doesn't throw an exception, it worked (see test below)
```
Justintime50, line 170: This should be a note in a docblock with the first sentence describing the test itself.

test/EasyPost/mocking/MockingUtility.php
```diff
@@ -0,0 +1,31 @@
...
+    public function __construct(MockRequest ...$mockRequests)
+    {
+        $this->mockRequests = $mockRequests;
+    }
+
+    public function findMatchingMockRequest($method, $url)
+    {
+        foreach ($this->mockRequests as $mockRequest) {
+            $methodMatches = $mockRequest->matchRule->method === '' || $mockRequest->matchRule->method === $method;
+            $urlMatches = $mockRequest->matchRule->urlRegexPattern == '' || preg_match($mockRequest->matchRule->urlRegexPattern, $url) >= 1;  // limit to exactly one match?
```
Justintime50, line 23: Why a question mark?

Justintime50 (changes requested):

Justintime50:
Please also provide a screenshot showing that test coverage includes these lines of code. I'm glad the tests pass, but if the tests don't actually run the code in question then this adds no benefit.

lib/EasyPost/EasyPostClient.php
nwithan8: It's an object, not a boolean

lib/EasyPost/EasyPostClient.php
nwithan8: It's an object, not a boolean

test/EasyPost/BillingTest.php
nwithan8: If the HTTP request (a real one) were to throw an exception. Crafting the test like it's making real requests (see webhook delete test)

test/EasyPost/BillingTest.php
nwithan8: This is used to set the global client.
Some tests however need to set up a local client with specific mock requests override, which is what this function is for.

test/EasyPost/BillingTest.php
nwithan8: Alright, I figured the function names were self-explanatory.

nwithan8:
> Please also provide a screenshot showing that test coverage includes these lines of code. I'm glad the tests pass, but if the tests don't actually run the code in question then this adds no benefit.

`make coverage`:
image

Justintime50:
This screenshot still doesn't show the percentage coverage and if it went up from the previous number. Either find the coverage percentage on the CLI and ensure it's at least a full point or more above the minimum or open the coverage HTML report and screenshot/verify that.

Justintime50 (dismissed):
Please bump the minimum coverage number found in the Makefile

Justintime50 (approved):

jchen293 (approved):

PR #248, opened 2022-12-16
fix: recursion issues with EasyPostClient, deprecation warnings

Justintime50:
# Description

Fixes the recursion problems of the EasyPostClient object by only instantiating the service when needed instead of when a client is created and passing those infinitely down. 

Also fixes a few deprecation warnings that were missed previously due to suppressed output.

# Testing

Dumped an EasyPostClient object to ensure the recursion was fixed, re-recorded a cassette locally to ensure that services/functions still get called as expected

# Pull Request Type

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -9,7 +9,7 @@ clean:
 ## coverage - Runs the test suite and generates a coverage report
 coverage:
 	composer coverage
-	bin/coverage-check build/logs/clover.xml 80 --only-percentage
+	bin/coverage-check build/logs/clover.xml 81 --only-percentage
 
 ## docs - Generate documentation for the library
 docs:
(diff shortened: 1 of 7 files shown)


nwithan8 (dismissed):

nwithan8:
> only instantiating the service when needed instead of when a client is created

This goes counter to how it was designed in Java (does match how C# works, but that was something we wanted to revert).

Justintime50:
True. For PHP, this is the only way to remove the recursion, it ran WAY faster doing so. To the user, they'll never know. I think either approach is acceptable depending on the language. Recursion may not be an issue in Java due to language differences.

Ultimately what we *really* want to avoid is requiring the user to instantiate the services. Whether that's on the fly or at object creation, as long as we handle it for them, I think we're good.

jchen293 (dismissed):

nwithan8 (approved):

jchen293 (approved):

nwithan8:
Sorry, my mocking merge caused a conflict here.

PR #254, opened 2023-01-11
fix: safer class names by using class builtins instead of strings

Justintime50:
# Description

After staring at this code for awhile, I realized that we should be using builtin class selectors instead of loose strings. This provides a much safer binding between services, tests, and models. It also helps IDE's know what in the world the class points to.

# Testing

Tests pass with changes

# Pull Request Type

### DIFF
--- a/lib/EasyPost/Address.php
+++ b/lib/EasyPost/Address.php
@@ -2,8 +2,6 @@
 
 namespace EasyPost;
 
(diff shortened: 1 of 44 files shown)


nwithan8 (approved):
michael-scott-the-office

jchen293 (approved):

PR #257, opened 2023-01-13
feat: Add functions to retrieve event payload(s)

nwithan8:
# Description

- Add function to retrieve payload/payloads for an event
- Add new Payload class

# Testing

- New unit tests, cassettes

# Pull Request Type

### DIFF
--- a/lib/EasyPost/Service/EventService.php
+++ b/lib/EasyPost/Service/EventService.php
@@ -2,6 +2,9 @@
 
 namespace EasyPost\Service;
 
+use EasyPost\Http\Requestor;
+use EasyPost\Util\InternalUtil;
+
 /**
  * Event service containing all the logic to make API calls.
  */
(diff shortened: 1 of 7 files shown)


test/EasyPost/EventTest.php
```diff
@@ -93,4 +94,94 @@ public function testReceiveNoInput()
...
+            'url' => Fixture::webhookUrl(),
+        ]);
+
+        // Create a batch to trigger an event.
+        self::$client->batch->create([
+            'shipments' => [Fixture::basicShipment()],
+        ]);
+
+        // Wait for the event to be created.
+        sleep(5);
```
Justintime50, line 116: Only on record

test/EasyPost/EventTest.php
```diff
@@ -93,4 +94,94 @@ public function testReceiveNoInput()
...
+
+        // Retrieve all events and extract the newest one.
```
Justintime50, line 157: The comments in this (and your other) PRs are too verbose. We can see *what* is happening by reading the code. Comments should only exist to explain *why* we need something. The only real two places this should be needed in this test and the other is to describe why we are sleeping and why we need this very specific payload ID

Justintime50 (dismissed):

lib/EasyPost/Service/EventService.php
```diff
@@ -28,4 +31,35 @@ public function all($params = null)
...
+    /**
+     * Retrieve a payload for an event.
+     *
+     * @param string $id The event id
+     * @param string $payloadId The payload id
+     * @return mixed
+     */
+    public function retrievePayload($id, $payloadId)
+    {
+        $url = $this->instanceUrl(self::serviceModelClassName(self::class), $id) . '/payloads' . '/' . $payloadId;
```
Justintime50, line 59: No need to split two strings up when they can be consolidated

```suggestion
        $url = $this->instanceUrl(self::serviceModelClassName(self::class), $id) . '/payloads/' . $payloadId;
```

Justintime50 (approved):

jchen293 (approved):

PR #262, opened 2023-02-01
add stateless rating

jchen293:
# Description

Add `retrieveStatelessRates` function and helper function `getLowestStatlessRate`
# Testing

Add two unit tests

# Pull Request Type


lib/EasyPost/Service/BetaRateService.php
```diff
@@ -0,0 +1,34 @@
...
+    public function retrieveStatelessRates($params)
+    {
+        $wrappedParams = [
+            'shipment' => $params,
+        ];
+
+        $response = Requestor::request($this->client, 'post', '/rates', $wrappedParams, true)['rates'];
+
+        foreach ($response as &$key) {
+            $key['object'] = 'StatelessRate';
```
nwithan8, line 28: *shutters*

nwithan8 (dismissed):

lib/EasyPost/Service/BetaRateService.php
jchen293: what?

nwithan8: Just, I continue to hate the way we automagically determine what object to deserialize the JSON into via its `object` key, rather than being explicit about it.

Nothing wrong with this PR.

jchen293: Yea... Unfortunately, our API returns `rate` as the object name for Rate and StatelessRate. Had to do some tweaks to make them separate objects.

Justintime50: They aren't separate objects. They are one in the same - a stateless rate just doesn't have an ID.

We should not be creating a whole new object for this since they are the same. Instead, these should be rate objects because that's how they'll be returned from the API. Manipulating the data in the library isn't the right approach since it will not match the docs or API contract.

Justintime50 (changes requested):

lib/EasyPost/Service/BetaRateService.php
nwithan8: We are creating new "StatelessRate" objects in C#, Java and Go, which have all the same properties of a "Rate" minus an ID, created_at and updated_at property...

jchen293: In other CLs, we created the `statelessRate` class, are we making an exception in PHP lib?

jchen293: After the internal team convo, we have decided to use `rate` for stateless rate for now

lib/EasyPost/Util/Util.php
```diff
@@ -132,4 +132,75 @@ public static function receiveEvent($rawInput = null)
...
+     * Get the lowest rate of stateless rates.
+     * To exclude a carrier or service, prepend the string with `!`.
+     *
+     * @param array statelessRates
+     * @param array $carriers
+     * @param array $services
+     * @return Rate
+     * @throws \EasyPost\Exception\EasyPostException
+     */
+    public static function getLowestStatlessRate($statelessRates, $carriers = [], $services = [])
```
Justintime50, line 146:
```suggestion
    public static function getLowestStatelessRate($statelessRates, $carriers = [], $services = [])
```

lib/EasyPost/Util/Util.php
```diff
@@ -132,4 +132,75 @@ public static function receiveEvent($rawInput = null)
             throw new EncodingException(Constants::DECODE_WEBHOOK_ERROR);
         }
     }
+
+    /**
+     * Get the lowest rate of stateless rates.
```
Justintime50, line 137:
```suggestion
     * Get the lowest stateless rate.
```

Justintime50 (changes requested):

nwithan8 (approved):

Justintime50 (approved):

PR #265, opened 2023-02-15
feat: add model properties and deserialize into each property

Justintime50:
# Description

This is a POC to show how we could add class properties to each model and deserialize directly into those properties.

We **cannot** move on this PR until we drop support for PHP 7.4 since we must have union typing available to us which is only supported on PHP 8.0+. Once that is possible, we could do something like this.

# Testing

# Pull Request Type


Justintime50:
This serves only as a POC and is published for posterity. We'll revisit this some months from now when we decide to drop PHP 7.4 support.

PR #266, opened 2023-02-15
fix: object mapping, required return types of PHP magic functions

Justintime50:
# Description

The object mapping list was missing `Brand` and had a few extra items in the list that never come back in an `object` field via a response. Although they are local classes in this lib, the constant list they were a part of is to lookup a value based on what we get back from the API, which in these cases would never happen. Removed to tighten up the possible options.

# Testing

- Tests pass

# Pull Request Type

### DIFF
--- a/lib/EasyPost/EasyPostObject.php
+++ b/lib/EasyPost/EasyPostObject.php
@@ -152,9 +152,9 @@ class EasyPostObject implements \ArrayAccess, \Iterator
      *
      * @param string $k
      * @param mixed $v
-     * @return void
      */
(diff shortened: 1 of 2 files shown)


nwithan8 (approved):
This witchcraft LGTM.

Justintime50:
@nwithan8 sorry for the need of another review. Turns out that even still with supporting PHP 7.4 we can't do the proper return types here. Removed those for now. I thought that dropping 7.3 was enough but I was mistaken.

nwithan8 (approved):

Issue #272, opened 2023-03-02
[Bug]: Composer lockfile added to repo


hotmeteor:
### Software Version

^6.0

### Language Version

PHP 8.2

### Operating System

macOS

### What happened?

I've forked this repo to add some contributions, but upon running `composer install` I noticed that a lockfile existed. I'm not clear why a lockfile would exist in a package repo like this? Can we remove it?

### What was expected?

`composer.lock` file not existing in package repo

Justintime50:
Hey @hotmeteor, thanks for the issue. Can you help us understand why removal of the `composer.lock` file is necessary? It helps keep everyones dependencies the same during development.

hotmeteor:
Sure. If you're writing a package for public consumption you don't want to have a lockfile, since everyone's system is going to be slightly different and the goal is for the package to be as widely compatible as you're willing to support.

For example, you might be cool to support PHP 7.4 => 8.2 and beyond. If that's the case, then you'll allow for multiple versions of your different dependencies to be installed, depending on the PHP version your consumer is using.

Here's a sample of what I mean from another package I'm using:

```json
 "require": {
      "php": "^8.1",
      "illuminate/support": "^9.0|^10.0"
 },
  "require-dev": {
      "nunomaduro/collision": "^6.0|^7.0",
      "orchestra/testbench": "^7.0|^8.0",
      "phpunit/phpunit": "^9.5.8|^10.0",
      "stripe/stripe-php": "^9.2"
 },
```
This package supports both 8.1 and 8.2, as the various dependency versions allow. You could imagine how supporting even more PHP versions would be possible (as long as your package code works across all of those PHP versions).

By **not including** the lockfile you're allowing your users to install the package without any conflicts to versions between the code and their environment.

As it is today, by **including** the lockfile you're assuming everyone in the world using your package has the exact same env setup as you.

Make sense?

hotmeteor:
Note: this position isn't the same if it's a "private" repo for something you're building with a team. Then, the expectation is that everyone has more or less the same env setup and you want things to be locked in as consistently as possible.

But packages, not so much. Go take a look at any popular PHP package or SDK, there won't be a lockfile.

PR #273, opened 2023-03-02
#272 Remove composer.lock from repo

hotmeteor:
# Description

This removes `composer.lock` from the repo since it's atypical to include it in a public package.

# Testing

In ran `composer install` after removing the lockfile and wasn't warned about the lockfile being out of sync.

# Pull Request Type


nwithan8 (approved):
Thanks for catching this!

jchen293 (approved):

Justintime50 (approved):
I fear I may have added this long ago with the misunderstanding between service/app and library.

You are correct in that libraries should not commit their composer.lock files. Thanks for putting up a PR that addresses this!

Issue #286, opened 2023-04-10
[Bug]: Using null as Exception message is deprecated in PHP 8.1


Chekote:
### Software Version

6.4.0

### Language Version

8.1

### Operating System

N/A

### What happened?

Without passing any arguments, instantiate any of the project's custom Exceptions that use null as a default value for $message.

PHP throws a deprecation notice.

### What was expected?

The new Exception object should be created.

### Sample Code

```inc
new \EasyPost\Exception\General\EasyPostException();
```

### Proposed fix:
Set the default to an empty string instead.

PR #287, opened 2023-04-10
Fix deprecation warnings for null Exception message

Chekote:
Fixes EasyPost/easypost-php#286

# Description
Fixes deprecation warnings when using PHP 8.1. Please see linked issue for details.

> Passing null to non-nullable internal function parameters is deprecated.

Source: https://www.php.net/releases/8.1/en.php

# Testing
No additional testing is required. This does not add any new features or functionality. The existing test suite should catch any problems.

# Pull Request Type

### DIFF
--- a/lib/EasyPost/Exception/Api/ApiException.php
+++ b/lib/EasyPost/Exception/Api/ApiException.php
@@ -25,7 +25,7 @@ class ApiException extends EasyPostException
      * @param int $httpStatus
      * @param string $httpBody
      */
-    public function __construct($message = null, $httpStatus = null, $httpBody = '')
+    public function __construct($message = '', $httpStatus = null, $httpBody = '')
     {
         parent::__construct($message);
         $this->httpStatus = $httpStatus;
--- a/lib/EasyPost/Exception/General/EasyPostException.php
+++ b/lib/EasyPost/Exception/General/EasyPostException.php
@@ -9,7 +9,7 @@ class EasyPostException extends \Exception
      *
      * @param string $message
      */
-    public function __construct($message = null)
+    public function __construct($message = '')
     {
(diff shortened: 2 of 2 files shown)


nwithan8:
Thanks for the submission! Looks good to me!

nwithan8 (approved):

Justintime50 (commented):
Thanks for submitting this!

It's good, can we do better? Is there ever a need to instantiate an error without a message? Can we instead remove the fallbacks of null/empty string and have message be more explicitly required? I can't foresee us ever throwing an error without stating what the error actually is.

Chekote:
Hi @Justintime50 

Thanks for the feedback.

While I do agree with your feedback, I feel like that is a separate concern that should be addressed within a different issue/PR. My reasoning is:

* Even without this PR, the objection you describe already exists in the codebase.
* This PR is not intended to make unrelated improvements to the behavior of the project.
* This PR resolves the problem of PHP 8.1 compatibility without introducing further technical debt.
* Making the message required would be a breaking change to the method signature. Fixing PHP 8.1 compatibility without introducing a breaking change would be preferable.

In the interests of agile development and resolving the problem outlined in the issue quickly and with minimal disruption to the package users (see my point above on breaking changes), I suggest changes in this PR be limited to the scope of the issue at hand and address your suggestion separately in the next major release.

Justintime50:
I'll accept this change because it addresses the deprecation notice; however, I don't see this as a breaking change since the producer of these exceptions should really only be this library. Making the `message` required shouldn't affect users and we already pass a message to every exception in the library so it wouldn't change behavior on our end.

Are you currently creating `ApiException`s or `EasyPostException`s in your integration?

PR #300, opened 2023-07-17
feat: adds http request and response hooks

Justintime50:
# Description

- Adds new `RequestHook` and `ResponseHook` events. (un)subscribe to them with the new `subscribeToRequestHook`, `subscribeToResponseHook`, `unsubscribeFromRequestHook`, or `unsubscribeFromResponseHook` methods of an `EasyPostClient`

# Testing

Adds a new set of tests that subscribe to the hooks and assert details only exposed via the hooks (not the traditional deserialized objects)

Adds a test to ensure that unsubscribing works as intended (will raise an exception if still subscribed), A/B tested this to ensure it works.

# Pull Request Type

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,26 @@ $boughtShipment = $client->shipment->buy($shipment->id, $shipment->lowestRate())
 echo $boughtShipment;
 ```
 
+### HTTP Hooks
+
+Users can subscribe to HTTP requests and responses via the `RequestHook` and `ResponseHook` objects. To do so, pass a function to the `subscribeToRequestHook` or `subscribeToResponseHook` methods of an `EasyPostClient` object:
+
+```php
+function customFunction($args)
+{
+    // Pass your code here, data about the request/response is contained within `$args`.
+    echo "Received a request with the status code of: " . $args['http_status'];
+}
+
+$client = new \EasyPost\EasyPostClient(getenv('EASYPOST_API_KEY'));
+
+$client->subscribeToResponseHook('customFunction');
+
+// Make your API calls here, your customFunction will trigger once a response is received
+```
+
+You can also unsubscribe your functions in a similar manner by using the `unsubscribeFromRequestHook` and `unsubscribeFromResponseHook` methods of a client object.
+
 ## Documentation
 
 API documentation can be found at: <https://easypost.com/docs/api>.
--- a/lib/EasyPost/EasyPostClient.php
+++ b/lib/EasyPost/EasyPostClient.php
@@ -5,6 +5,8 @@ namespace EasyPost;
 use EasyPost\Constant\Constants;
 use EasyPost\Exception\General\EasyPostException;
 use EasyPost\Exception\General\MissingParameterException;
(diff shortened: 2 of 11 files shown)


lib/EasyPost/Hook/EventHook.php
```diff
@@ -0,0 +1,33 @@
...
+ * The parent event that occurs when a hook is triggered.
+ */
+class EventHook
+{
+    private $eventHandlers = [];
+
+    public function __invoke(...$args)
+    {
+        foreach ($this->eventHandlers as $eventHandler) {
+            $eventHandler(...$args);
```
nwithan8, line 15: Calling a function by its name. God, I had dynamic languages.

nwithan8 (approved):

lib/EasyPost/Hook/EventHook.php
Justintime50: This is definitely a phpism... and I hate it too 😂

README.md
```diff
@@ -63,6 +63,26 @@ $boughtShipment = $client->shipment->buy($shipment->id, $shipment->lowestRate())
...
+}
+
+$client = new \EasyPost\EasyPostClient(getenv('EASYPOST_API_KEY'));
+
+$client->subscribeToResponseHook('customFunction');
+
+// Make your API calls here, your customFunction will trigger once a response is received
+```
+
+You can also unsubscribe your functions in a similar manner by using the `unsubscribeFromRequestHook` and `unsubscribeFromResponseHook` methods of a client object.
```
jchen293, line 84: I would provide an unsubscribe example.

lib/EasyPost/Hook/EventHook.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+namespace EasyPost\Hook;
+
+/**
+ * The parent event that occurs when a hook is triggered.
+ */
+class EventHook
```
jchen293, line 8: Need docstrings for these functions in this class

lib/EasyPost/EasyPostClient.php
```diff
@@ -65,6 +67,8 @@ class EasyPostClient extends BaseService
     private $timeout;
     private $apiBase;
     private $mockingUtility;
+    public $requestEvent;
+    public $responseEvent;
```
jchen293, line 71: Ideally, we should use getters for these two attributes

jchen293 (commented):
lgtm, left a few non-blocking comments

jchen293 (approved):
I meant to approve this with my previous comments

README.md
Justintime50: I don't think it's different enough to warrant an example. It states you can just swap in the name of the function here

lib/EasyPost/EasyPostClient.php
Justintime50: Potentially; however, these attributes aren't used in the same manner a normal attribute would be. Additionally they aren't meant for public consumption so I don't feel bad keeping them this way.

PR #311, opened 2023-10-04
fix: reset timezone to original (closes #310)

### LINKED ISSUE #310: [Bug]: Timezone is changed and not changed back
### Software Version

6.8.1

### Language Version

8.1.11

### Operating System

Ubuntu

### What happened?

Your client is forcing the timezone to UTC to get a UTC timestamp in microseconds, but you're not switching it back. Line [186](https://github.com/EasyPost/easypost-php/blob/5655532be2373865cdca567e770ebf6c44f83155/lib/EasyPost/Http/Requestor.php#L186) in lib/EasyPost/Http/Requestor.php.

### What was expected?

As of PHP 7.1 the built-in [DateTime class](https://www.php.net/manual/en/class.datetime.php) supports microseconds. You can create a new DateTime instance, change the timezone of the instance to UTC, then format it as a timestamp. You could also store the current timezone, set it to UTC, call microtime(), then change the timezone back.

### Sample Code

_No response_

### Relevant logs

_No response_

Justintime50:
# Description

- Fixes a bug that changed the timezone to UTC on each request without resetting it to the original timezone in use (closes #310)

# Testing

Ran these added commands in the repl to make sure these work as expected.

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next Release
 
 - Migrate API Key-related functions from `user` service to `api_keys` service
+- Fixes a bug that changed the timezone to UTC on each request without resetting it to the original timezone in use (closes #310)
 
 ## v6.8.1 (2023-09-05)
 
--- a/lib/EasyPost/Http/Requestor.php
+++ b/lib/EasyPost/Http/Requestor.php
@@ -183,6 +183,7 @@ class Requestor
         ];
 
         $requestUuid = uniqid();
+        $originalTimezone = date_default_timezone_get();
         date_default_timezone_set('UTC');
         $requestTimestamp = microtime(true);
         ($client->requestEvent)([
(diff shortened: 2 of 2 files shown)


nwithan8 (approved):

dcaballeroc (approved):
A test might come in handy

Justintime50:
> A test might come in handy

I tested in the repl before pushing up, troubles with a test like this is we can't guarantee timezones of whatever machine it'll run on. Depending on CI or someones local machine, default timezones may differ and assertions wouldn't necessarily match

jchen293 (approved):

dcaballeroc:
> > A test might come in handy
> 
> I tested in the repl before pushing up, troubles with a test like this is we can't guarantee timezones of whatever machine it'll run on. Depending on CI or someones local machine, default timezones may differ and assertions wouldn't necessarily match

Timezones: not even once.

PR #312, opened 2023-10-04
fix: pin docs binary in Makefile

Justintime50:
# Description

Pins the docs binary. The docs binary had breaking changes in a non major release and when pulled in, the command we have doesn't work. Also the most recent version doesn't support PHP 7 which we still do

# Testing

Generated docs locally

# Pull Request Type

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ coverage:
 
 ## docs - Generate documentation for the library
 docs:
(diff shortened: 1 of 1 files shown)


dcaballeroc (approved):
It's a pin

jchen293 (approved):

PR #316, opened 2023-11-21
fix: pagination errors with user-supplied params

Justintime50:
# Description

- Fixes a bug where the original filtering criteria of `all` calls wasn't passed along to `getNextPage` calls. Now, these are persisted via a `_params` key on response objects locally
- Fixes error handling in `getNextPage` tests

# Testing

Now test for existence of `_params`

# Pull Request Type


lib/EasyPost/Service/ReportService.php
```diff
@@ -66,10 +67,19 @@ public function getNextPage($reports, $pageSize = null)
             'page_size' => $pageSize,
             'before_id' => $reportArray[count($reportArray) - 1]['id']
         ];
-        $url = self::reportUrl($reports->type);
+
+        if (isset($userParams)) {
+            $params = array_merge($params, $userParams);
+        }
+
+        $url = self::reportUrl($reportArray[0]->object);
```
nwithan8, line 75: **question:** ‏ Can you explain how this `reportUrl` function works? What is `object` in this situation?

nwithan8 (changes requested):

lib/EasyPost/Service/ReportService.php
Justintime50: object is the type of report. The code before wouldn't actually pass in anything it could use haha. This will pass in something like `ShipmentReport` and will turn it into `shipment`

dcaballeroc (commented):
Update pertinent cassettes as well?

dcaballeroc (approved):

nwithan8 (approved):

PR #325, opened 2023-11-30
phpstan

Justintime50:
# Description

This PR introduces PHPStan which is a static analysis tool. Think of it like `mypy` from Python but with additional tools built in for syntax checking in addition to type checking. This found SO many issues that needed fixing and will help enforce proper type hints.

# Testing

There are 3 additional levels of scrutiny we could use in this tool but it was getting overwhelming and I stopped here. There are some bigger changes needed to correct levels 7-9 so we'll revisit later.

# Pull Request Type


dcaballeroc (approved):
Nice!

Justintime50:
> PHP with actual type hints, my IDE thanks you.

I too am quite stoked

PR #334, opened 2024-03-26
[FEAT] Beta `paginated_api_keys` function

nwithan8:
# Description

- Adds new function `paginated_api_keys` to `User` to return a paginated list of API keys, rather than unpaginated from `api_keys`

**NOTE:** This beta function exists in the GA `User` class rather than in a separate `Beta\User` class because this method is an instance method. It would only be accessible on `Beta\User` objects, which wouldn't have any other of the existing functions that exist on the `User` object. They would either have to be entirely re-implemented or a converter between `User` and `Beta\User` would have to be introduced; both strategies would add extra overhead and require changes to the end-user's workflow.

More recent major versions of our library have effectively removed this issue by our re-architecture into services that make it easier to separate beta functions from GA functions. For now, a docstring has been added warning users that this new function is in beta.

# Testing

- Add unit test to confirm/validate data coming back from new function/endpoint, test pagination capabilities
- Add cassette

# Pull Request Type

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # CHANGELOG
 
+## Next Release
+
+- [ADDED] Beta `paginated_api_keys` functionality to the `User` class
+  - Retrieves a paginated list of API keys for the authenticated user, versus `api_keys`, which returns an unpaginated list
+
 ## v5.8.0 (2022-12-01)
 
 - [ADDED] Carrier account creation routes requests correctly for carriers that require custom workflows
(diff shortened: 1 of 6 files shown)


nwithan8:
CI is failing because security scan is failing because it's trying to install the latest version of `roave/security-advisories` which requires a newer version of PHP.

`roave/security-advisories` [does not offer](https://github.com/Roave/SecurityAdvisories?tab=readme-ov-file#stability) any tagged version of its library other than `dev-latest`, so we can't "pin an older version" to overcome this issue.

Justintime50 (approved):

jchen293 (approved):

PR #341, opened 2024-07-09
Update CarrierService update and create method to account UPS endpoints

jchen293:
# Description

Update CarrierService update and create method to account UPS endpoints
# Testing

Add unit tests

# Pull Request Type

### DIFF
--- a/lib/EasyPost/Constant/Constants.php
+++ b/lib/EasyPost/Constant/Constants.php
@@ -20,7 +20,7 @@ abstract class Constants
         'FedexSmartpostAccount'
     ];
 
-    const UPS_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS = [
+    const UPS_OAUTH_ACCOUNT_TYPES = [
         'UpsAccount',
         'UpsMailInnovationsAccount',
         'UpsSurepostAccount'
--- a/lib/EasyPost/Constant/Constants.php
+++ b/lib/EasyPost/Constant/Constants.php
@@ -17,8 +17,13 @@ abstract class Constants
     // Validation
     const CARRIER_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS = [
         'FedexAccount',
-        'UpsAccount',
         'FedexSmartpostAccount'
+    ];
+
+    const UPS_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS = [
+        'UpsAccount',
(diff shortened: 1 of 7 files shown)


nwithan8 (approved):

lib/EasyPost/Constant/Constants.php
```diff
@@ -17,10 +17,15 @@ abstract class Constants
     // Validation
     const CARRIER_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS = [
         'FedexAccount',
-        'UpsAccount',
         'FedexSmartpostAccount'
     ];
 
+    const UPS_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS = [
```
Justintime50, line 23: suggestion: Prefer this called something like `UPS_OAUTH_ACCOUNT_TYPES` to indicate how they're different than the normal custom workflows

lib/EasyPost/Service/CarrierAccountService.php
```diff
@@ -43,7 +43,16 @@ public function all(mixed $params = null): mixed
...
     public function update(string $id, mixed $params): mixed
     {
-        return self::updateResource(self::serviceModelClassName(self::class), $id, $params);
+        $carrierAccount = self::retrieve($id);
+        $carrierAccountType = $carrierAccount['type'];
+        if (in_array($carrierAccountType, Constants::UPS_ACCOUNT_TYPES_WITH_CUSTOM_WORKFLOWS, true)) {
+            $updateUrl = 'upsOauthRegistration';
+            $params = [self::selectTopLayerKey($carrierAccountType) => $params];
+        } else {
+            $updateUrl = 'carrierAccount';
```
Justintime50, line 52: These are not URLs and should not be named as such. They become a URL later but are actually class names here. Also, wouldn't they be title cased?

```suggestion
            $className = 'UpsOauthRegistration';
            $params = [self::selectTopLayerKey($carrierAccountType) => $params];
        } else {
            $className = 'CarrierAccount';
```

jchen293: I don't think it's case sensitive, but I have uppercased the first character

Justintime50 (approved):

lib/EasyPost/Service/CarrierAccountService.php
Justintime50: Class names must be capitalized, otherwise it's not a true class name.

PR #393, opened 2026-08-26
fix: preserves caller-provided plain PHP objects in request params (closes #392)

### LINKED ISSUE #392: 8.8.2: empty PHP arrays now reach the API as JSON arrays where objects are expected — request rejected with no pointer to the cause
## Summary

The v8.8.2 change ("Empty arrays and objects are no longer stripped when passed as params in the request") has a PHP-specific consequence that's easy to hit and hard to trace back to the SDK: an empty PHP array now reaches the API as a JSON array (`[]`) in positions where the API expects an object, and the request is rejected.

We're not asking you to revert — forwarding what the caller passed is a reasonable position. We're raising this because the failure mode is hard to diagnose, and other consumers will likely hit it as 8.8.2 propagates.

## Details

PHP has no way to distinguish an empty list from an empty map — `[]` is the only natural empty value, and `json_encode` picks the array reading. Code like this is common:

```php
$options = [];

if ($needsDdp) {
    $options['incoterm'] = 'DDP';
}

$shipment = $client->shipment->create([
    // ...
    'options' => $options,   // [] when no branch fired
]);
```

Before 8.8.2 the SDK stripped the empty `options`, so this worked. On 8.8.2 the API receives `"options": []` and rejects the shipment with:

```
The provided options are not valid.
```

Nothing in that message points at the SDK, the upgrade, or the empty array — for us it surfaced only as runtime API errors on previously working code after a routine patch-level composer update.

We verified against the live API that the rejection is purely about JSON type: the same shipment with `"options": {}` (raw request; the SDK can't produce an empty object from PHP) returns 201 with rates, `"options": []` returns 422, and omitting the key returns 201.

## Suggestions

Any of these would help, roughly in order of value:

1. **Coerce empty arrays to `{}` for params documented as objects** (`options`, `customs_info`, …). Since `{}` is accepted by the API, this fully restores the conditional-build pattern above — and there's currently no way to send an empty object through the SDK at all (`Requestor::encodeObjects` handles arrays and `EasypostObject`s, so a caller-supplied `stdClass` doesn't survive encoding either).
2. **Add a compatibility warning to the 8.8.2 changelog entry and UPGRADE guide** noting that previously-stripped empty params are now sent, and that object-typed params passed as `[]` will be rejected by the API — so people who hit "The provided options are not valid." can find their way here by searching.
3. Failing either, a note in the README's request-params section about the `[]` vs `{}` ambiguity.

Happy to open a PR for (1) or (2) if that's welcome.

## Environment

- easypost-php 8.8.2 (upgraded from 8.7.0)
- PHP 8.4
- Observed with `shipment->create` when `options` is an empty array; any object-typed param passed as `[]` should reproduce.

Justintime50:
# Description
- Preserves caller-provided plain PHP objects in request params so `(object) []` is sent as an empty JSON object (`{}`) instead of being stringified (closes #392)

# Testing

Added new unit tests that ensure backwards compatibility along with the change of standard objects being encoded properly with the fix.

# Pull Request Type

### DIFF
--- a/lib/EasyPost/Http/Requestor.php
+++ b/lib/EasyPost/Http/Requestor.php
@@ -65,13 +65,13 @@ class Requestor
      * Encodes an EasyPost object and prepares the data for the request.
      *
      * @param mixed $data
-     * @return array<mixed>|string
+     * @return mixed
      */
-    private static function encodeObjects(mixed $data): array|string
+    private static function encodeObjects(mixed $data): mixed
     {
         if (is_null($data)) {
             return [];
-        } elseif ($data instanceof EasypostObject) {
+        } elseif ($data instanceof EasyPostObject) {
             return ['id' => self::utf8($data->id)]; // @phpstan-ignore-line
         } elseif ($data === true) {
             return 'true';
@@ -86,6 +86,15 @@ class Requestor
             }
 
             return $resource;
+        } elseif ($data instanceof \stdClass) {
+            $resource = [];
+            foreach (get_object_vars($data) as $k => $v) {
+                if (!is_null($v) and ($v !== '')) {
+                    $resource[$k] = self::encodeObjects($v);
+                }
+            }
+
+            return (object)$resource;
         } else {
             return self::utf8(strval($data));
         }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # CHANGELOG
 
+## v8.8.3 (2026-08-26)
+
+- Preserves caller-provided plain PHP objects in request params so `(object) []` is sent as an empty JSON object (`{}`) instead of being stringified
+
 ## v8.8.2 (2026-08-11)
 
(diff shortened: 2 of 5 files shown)


lib/EasyPost/Http/Requestor.php
```diff
@@ -86,6 +86,15 @@ private static function encodeObjects(mixed $data): array|string
...
+            $resource = [];
+            foreach (get_object_vars($data) as $k => $v) {
+                if (!is_null($v) and ($v !== '')) {
+                    $resource[$k] = self::encodeObjects($v);
+                }
+            }
+
+            return (object)$resource;
         } else {
             return self::utf8(strval($data));
```
eschalburg-easypost, line 99: Won't this still stringify the response?

Justintime50: yes though it's not in the elseif block for this specific path having issues (the newly added code)

eschalburg-easypost (approved):

